PR #2, opened 2022-09-12
Feat/wip

pgautier404:
MVP for PHP SDK.


README.md
```diff
@@ -1,5 +1,29 @@
...
+
+- A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
+- At least PHP 7
+- The grpc PHP extension. See the [gRPC docs](https://github.com/grpc/grpc/blob/v1.46.3/src/php/README.md) section on installing the extension.
+
+## Using Momento
+
+Run `composer update` from the root of the repository to install the necessary prerequisites.
+
+Then check out full working code in `example.php` at the root of this repository!
```
eaddingtonwhite, line 28: Probably need add section here for updating local protos for now.  Similar to https://github.com/momentohq/client-sdk-go#updating-grpc-protos

composer.json
```diff
@@ -1,12 +1,37 @@
...
+    },
+    "autoload-dev": {
+        "psr-4": { "Momento\\Tests\\": "tests/" }
+    },
     "require": {
-        "momentohq/client-types-php": "dev-main"
+        "php": ">=7.4",
+        "google/protobuf": "^3.21",
+        "grpc/grpc": "^1.42",
+        "ext-grpc": "*"
```
eaddingtonwhite, line 26: is it possible to keep all libraries pinned explicitly?

README.md
pgautier404: The protos are bundled in the SDK for right now, so that doesn't belong in the README at this point.

composer.json
pgautier404: Probably, but they're pinned in the composer.lock file already, so those versions will be what customers get until we regenerate it (now that I've corrected the README). I'd rather let these float in composer.json so we can receive updates without having to constantly monitor for them. I'm happy to discuss if you have any objection to that.

eaddingtonwhite: Ok, I generally prefer to keep pinned here and then have something like dependabot open PR's back in bumping versions. Just helps keep library upgrades self-contained and helps us merge them on our terms and never accidentally as part of another unrelated pr. We're a little inconsistent on this across repos though right now so lets maybe sync offline can come back to it.

README.md
eaddingtonwhite: I mean more like dev doc for like how to regenerate/update if need to say if we add new API. Doesn't need to be in README maybe contributing.md? Just from perspective of like if I had to come in and bump these as someone contributing to this SDK.

pgautier404: That process is in such flux (and only available to Momento internal) at this point that I'd rather put this off until we have something reasonable in place. Right now there's a `client-types-php` repo (that is going away in favor of a better solution when I have one) that has a manually triggered github action that produces a zipfile that you need to overwrite the protos in the SDK repository. Documenting the existing process isn't useful to the public at this point, but I'll definitely add a section for that when there's a sane process to document.

composer.json
cprice404: I'm okay with coming back to this later, doesn't have to be part of this PR, but I agree with Ellery - we should pin explicitly.

cprice404:
should I assume that everything in the `types` directory is generated and doesn't warrant thorough review?

example.php
```diff
@@ -0,0 +1,47 @@
...
+printBanner("*                      Momento Example Start                     *");
+$client = new SimpleCacheClient($MOMENTO_AUTH_TOKEN, $ITEM_DEFAULT_TTL_SECONDS);
+createCache($client, $CACHE_NAME);
+listCaches($client);
+print "Setting key $KEY to value $VALUE\n";
+$client->set($CACHE_NAME, $KEY, $VALUE);
+$response = $client->get($CACHE_NAME, $KEY);
+print "Look up status is: {$response->status()}\n";
+print "Look up value is: {$response->value()}\n";
+printBanner("*                       Momento Example End                      *");
```
cprice404, line 47: noice

composer.json
pgautier404: Dependabot is currently only checking versions on Momento libraries so wouldn't catch other library versions getting bumped as currently set up. I went ahead and pinned these in `composer.json` but we'll probably need to discuss this later.

pgautier404:
> should I assume that everything in the `types` directory is generated and doesn't warrant thorough review?

Yes, that's the generated code.

composer.json
cprice404: yep happy to discuss.

README.md
```diff
@@ -1,5 +1,29 @@
-# Momento client-types-php
+# client-sdk-php
 
-:warning: Pre-Alpha :warning:
+:warning: Experimental SDK :warning:
+
+PHP SDK for Momento is experimental and under active development. There could be non-backward compatible changes or
+removal in the future. Please be aware that you may need to update your source code with the current version of the SDK
+when its version gets upgraded.
+---
```
cprice404, line 8: not a blocker for this PR but let's get this merged:

https://github.com/momentohq/standards-and-practices/pull/19

and then let's switch this repo over to use it.

README.md
```diff
@@ -1,5 +1,29 @@
...
+
+<br/>
+
+# Getting Started :running:
+
+## Requirements
+
+- A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
+- At least PHP 7
+- The grpc PHP extension. See the [gRPC docs](https://github.com/grpc/grpc/blob/v1.46.3/src/php/README.md) section on installing the extension.
```
cprice404, line 22: i take it that composer can't do this?

README.md
```diff
@@ -1,5 +1,29 @@
...
+traditional caching solutions.
+
+<br/>
+
+# Getting Started :running:
+
+## Requirements
+
+- A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
+- At least PHP 7
```
cprice404, line 21: I'm a php supernoob - is this going to be an acceptable minimum version?  I have no idea how old or new that is :)

README.md
```diff
@@ -1,5 +1,29 @@
...
+- The grpc PHP extension. See the [gRPC docs](https://github.com/grpc/grpc/blob/v1.46.3/src/php/README.md) section on installing the extension.
+
+## Using Momento
+
+Run `composer install` from the root of the repository to install the necessary prerequisites.
```
cprice404, line 26: does `composer` come with the PHP install?  or do we need to tell folks how to install that?

README.md
```diff
@@ -1,5 +1,29 @@
...
+
+Then check out full working code in `example.php` at the root of this repository!
```
cprice404, line 28: nit, in subsequent PRs let's just move this into an examples subdir because there will be more than one over time

src/Cache/Errors/AlreadyExistsError.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+namespace Momento\Cache\Errors;
+
+class AlreadyExistsError extends MomentoServiceError {}
```
cprice404, line 5: nit - does PHP restrict to a single class per file?  Just curious if we could consolidate the error classes.

src/Cache/_ScsControlClient.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Momento\Cache;
```
cprice404, line 3: nit, for consideration in later PRs: does PHP give class names that start with an `_` special treatment?  (e.g. does it hide them from autocomplete in IDEs?)  if not, maybe we move these to a sub-namespace.

src/Cache/_ScsControlClient.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function listCaches(?string $nextToken=null): ListCachesResponse
+    {
+        try {
+            $request = new _ListCachesRequest();
+            $request->setNextToken($nextToken ? $nextToken : "");
+            $call = $this->grpcManager->client->ListCaches($request);
+            [$response, $status] = $call->wait();
+        } catch (\Exception $e) {
+            // TODO: error converter and exception-less handling
```
cprice404, line 70: good call, I think we just deal with this the new way out of the gate

cprice404 (approved):
left some questions / nits, no blockers

README.md
pgautier404: Correct, this is a compiled extension.

README.md
pgautier404: gRPC requires 7+ so, yes :-)

README.md
pgautier404: Great call, I'll add a link to instructions.

README.md
pgautier404: I'm totally planning on that, but the way composer works I need to get this PR merged before I can hook that up cleanly. Moving the script into it's own directory with it's own `composer.json` to grab the entire SDK is my next step for this after merge.

src/Cache/Errors/AlreadyExistsError.php
pgautier404: Haha, turns out it doesn't despite my initial confusion about the PHP autoloader. I figured out how to consolidate these but didn't want to delay delivering this to fix that. Consolidating these is definitely part of my planned cleanup though.

src/Cache/_ScsControlClient.php
pgautier404: Let's talk about this offline . . . Not sure I completely understand what you're getting at.

README.md
cprice404: okay but how recent is that?  Is it new enough for use in wordpress etc.?  Just want to understand what things will and won't be open to us

src/Cache/_ScsControlClient.php
cprice404: sure we can chat any time today

PR #4, opened 2022-09-13
fix: consolidate error classes

pgautier404:


eaddingtonwhite (approved):
Nice looks cleaner 🧼 nice work cutting down # of files

PR #5, opened 2022-09-14
fix: add tests and fix bugs

pgautier404:
Added tests and fixed some issues they uncovered.

### DIFF
--- a/src/Cache/SimpleCacheClient.php
+++ b/src/Cache/SimpleCacheClient.php
@@ -2,6 +2,8 @@
 
 namespace Momento\Cache;
 
+use Momento\Cache\Errors\InvalidArgumentError;
+
 class SimpleCacheClient
 {
 
@@ -11,7 +13,7 @@ class SimpleCacheClient
      * @param string $authToken: momento JWT
      * @param int $defaultTtlSeconds: Default Time to Live for the item in Cache
      */
-    function __construct(string $authToken, int $defaultTtlSeconds, int $dataClientOperationTimeoutMs=0)
+    function __construct(string $authToken, int $defaultTtlSeconds, ?int $dataClientOperationTimeoutMs=null)
     {
         $payload = $this->parseAuthToken($authToken);
         $this->controlClient = new _ScsControlClient($authToken, $payload["cp"]);
@@ -20,9 +22,21 @@ class SimpleCacheClient
         );
     }
 
+    private function throwBadAuthToken() {
+        throw new InvalidArgumentError('Invalid Auth token.');
+    }
+
     private function parseAuthToken(string $authToken) : array {
-        list($header, $payload, $signature) = explode (".", $authToken);
-        return json_decode(base64_decode($payload), true);
+        $exploded = explode (".", $authToken);
+        if (count($exploded) != 3) {
+            $this->throwBadAuthToken();
+        }
+        list($header, $payload, $signature) = $exploded;
+        $token = json_decode(base64_decode($payload), true);
+        if ($token === null) {
+            $this->throwBadAuthToken();
+        }
+        return $token;
     }
 
     public function createCache(string $cacheName) : CacheOperationTypes\CreateCacheResponse
@@ -48,5 +62,10 @@ class SimpleCacheClient
     public function get(string $cacheName, string $key) : CacheOperationTypes\CacheGetResponse
     {
         return $this->dataClient->get($cacheName, $key);
+    }
+
+    public function delete(string $cacheName, string $key) : CacheOperationTypes\CacheDeleteResponse
+    {
+        return $this->dataClient->delete($cacheName, $key);
     }
 }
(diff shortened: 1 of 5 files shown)


src/Cache/SimpleCacheClient.php
```diff
@@ -20,9 +22,21 @@ function __construct(string $authToken, int $defaultTtlSeconds, int $dataClientO
...
 
+    private function throwBadAuthToken() {
+        throw new InvalidArgumentError('Invalid Auth token.');
+    }
+
     private function parseAuthToken(string $authToken) : array {
-        list($header, $payload, $signature) = explode (".", $authToken);
-        return json_decode(base64_decode($payload), true);
+        $exploded = explode (".", $authToken);
+        if (count($exploded) != 3) {
```
eaddingtonwhite, line 31: I think this is ok for now if we want. Feels like getting pretty close to where we should just use something off the shelf like https://github.com/firebase/php-jwt instead. @pgautier404 could you just make a follow ticket to track this so we dont loose sight of it can come back and revisit.

src/Cache/_ScsDataClient.php
```diff
@@ -1,30 +1,34 @@
...
 use Momento\Utilities\_ErrorConverter;
 use function Momento\Utilities\validateTtl;
 use function Momento\Utilities\validateCacheName;
+use function Momento\Utilities\validateOperationTimeout;
 
 class _ScsDataClient
 {
 
-    private static int $DEFAULT_DEADLINE_SECONDS = 5;
+    private static int $DEFAULT_DEADLINE = 500000; // I believe this maps to 5 sec.
```
eaddingtonwhite, line 20: so is this supposed to be in milliseconds now? Shouldnt it be 5000ms = 5 seconds? this is 50 seconds right now? Maybe missing something though
Also could we add unit to variable name as well so like  `$DEFAULT_DEADLINE_MS` if it is in ms.

src/Cache/_ScsDataClient.php
```diff
@@ -70,4 +74,19 @@ public function get(string $cacheName, string $key) : CacheGetResponse
...
+    {
+        validateCacheName($cacheName);
+        try {
+            $deleteRequest = new _DeleteRequest();
+            $deleteRequest->setCacheKey($key);
+            $call = $this->grpcManager->client->Delete($deleteRequest, ["cache" => [$cacheName]], ["timeout" => $this->deadline_seconds]);
+            [$response, $status] = $call->wait();
+        } catch (Exception $e) {
+            throw $e;
+        }
```
eaddingtonwhite, line 87: whats going on here w/ try/catch and rethrow? can we just erase whole thing was this left over from testing something?

src/Utilities/_DataValidation.php
```diff
@@ -3,21 +3,33 @@
...
+    function validateCacheName(string $cacheName) : void
     {
         if (!$cacheName || !is_string($cacheName)) {
-            throw new InvalidArgumentException("Cache name must be a non-empty string");
+            throw new InvalidArgumentError("Cache name must be a non-empty string");
+        }
+    }
+}
+
+if (!function_exists('validateOperationTimeout')) {
```
eaddingtonwhite, line 26: why do we need these if wrappers around the functions like this? Sorry I think missed in previous pr. Is this normal php thing?

src/Cache/SimpleCacheClient.php
pgautier404: Agreed, I'll make a ticket. This approach was just for maximum speed of delivery, as there were a number of PHP JWT libs to sift through.

src/Cache/_ScsDataClient.php
pgautier404: It definitely isn't in milliseconds, but I'm having a heck of a time figuring out what the unit **is** supposed to be. My closest rational guess is actually microseconds (!?). I've changed the code here to make this all a little more explicit and added a TODO comment for further research when time allows.

src/Cache/_ScsDataClient.php
pgautier404: This is set up for our future error handling where we don't bubble up exceptions. I've added comments to make that clearer.

src/Utilities/_DataValidation.php
pgautier404: The resource I modeled this after used this pattern. I think it's possible the PHP autoloader may try to process this file multiple times, in which case this would avoid errors about functions already being declared.

eaddingtonwhite: image

eaddingtonwhite (approved):
@pgautier404 Nice work on tests ! Good to get this merged and can continue w/ follow up prs for open items!

src/Cache/_ScsDataClient.php
eaddingtonwhite: yeah looks like µs looking again I think ur right

PR #9, opened 2022-10-11
fix: remove offending action

pgautier404:
This fixes an error in the push-to-main github action.


cprice404 (approved):

poppoerika (approved):
Assume we can remove this step since we're already doing this in the CI workflow?

pgautier404:
> Assume we can remove this step since we're already doing this in the CI workflow?

It's not valid in this context and was causing the action to fail. It is hooked up in the CI workflow though.

PR #14, opened 2022-10-12
feat: replace authToken with authProvider

pgautier404:


src/Cache/SimpleCacheClient.php
```diff
@@ -2,43 +2,30 @@
...
+    function __construct(
+        ICredentialProvider $authProvider, int $defaultTtlSeconds, ?int $dataClientOperationTimeoutMs=null
+    ) {
+        $this->controlClient = new _ScsControlClient($authProvider->getAuthToken(), $authProvider->getControlEndpoint());
         $this->dataClient = new _ScsDataClient(
-            $authToken, $payload["c"], $defaultTtlSeconds, $dataClientOperationTimeoutMs
+            $authProvider->getAuthToken(),
+            $authProvider->getCacheEndpoint(),
+            $defaultTtlSeconds,
+            $dataClientOperationTimeoutMs
```
malandis, line 25: Is there a standard `TimeSpan`-like construct in PHP? Would be strong to avoid hoisting number conversions on the users

src/Auth/AuthUtils.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
cprice404, line 1: so i take it that in modern PHP you don't need a closing tag for this anymore?  TIL

src/Cache/SimpleCacheClient.php
cprice404: I'm +1 to this idea, but if such a thing exists in PHP, we can handle it as a separate ticket/PR.

cprice404 (approved):
plan

PR #15, opened 2022-10-13
feat: improved error UX

pgautier404:
Update error handing to conform to client SDK spec. Uses `isXXX()` methods such as `asHit`, `asError`, etc. to provide type-safe instances and informative IDE code completion.

### DIFF
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -2,25 +2,173 @@
 namespace Momento\Cache\CacheOperationTypes;
 
 use Cache_client\_GetResponse;
-use Cache_client\ECacheResult;
 use Control_client\_ListCachesResponse;
 use Cache_client\_SetResponse;
-use Momento\Cache\Errors\InternalServerError;
+use Momento\Cache\Errors\MomentoErrorCode;
+use Momento\Cache\Errors\SdkError;
 
-class CreateCacheResponse {}
+trait ErrorBody {
+    private SdkError $innerException;
+    private MomentoErrorCode $errorCode;
+    private string $message;
 
-class DeleteCacheResponse {}
+    public function __construct(SdkError $error)
+    {
+        parent::__construct();
+        $this->innerException = $error;
+        $this->message = "{$error->messageWrapper}: {$error->getMessage()}";
+        $this->errorCode = $error->errorCode;
+    }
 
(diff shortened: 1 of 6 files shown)


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -2,25 +2,173 @@
...
+
+abstract class CreateCacheResponse extends ResponseBase {
+
+    public function asSuccess() : CreateCacheResponseSuccess|null
+    {
+        if ($this->isSuccess())
+        {
+            return $this;
+        }
+        return null;
```
cprice404, line 81: question: did you consider the tradeoffs between `null` vs throwing an exception here?

On the one hand, if someone calls this incorrectly, if we throw an exception then we can give a nice clear message about what went wrong, vs. the null pointer exception that they are likely to get otherwise.

On the other hand, doing it this way probably allows people to write code like this if they want to:

```
if ($foo = client.Get('blah', 'blah')) {
  // handle happy path
} else {
  // handle sad path
}
```

without any try/catch shenanigans.

I don't want to hold up your progress so we should definitely timebox any consideration on this and just keep moving, but floating it because I expect that python/JS/PHP are going to end up looking pretty similar for this stuff since they don't have algebraic data types.

cprice404 (approved):
looks great, just one minor question to consider.  it's something we can punt until later though if we don't come to an immediate consensus.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
cprice404: @malandis I'm curious if you have any strong opinions on this.

pgautier404: My thinking was that we have made a contract with the user that our cache client will always return and not throw exceptions. I would be upset as a user if I had to jump through a relatively exotic syntax **and** wrap the whole thing in a try/catch. As you pointed out, this also allows a pseudo-pattern matching scheme using assignments in if statements. The null pointer exception should be pretty easy to diagnose since the return type signature of these methods is explicitly "response class or null". We'll obviously want to demonstrate the patterns we end up deeming best practices in the docs.

cprice404: That seems pretty compelling.  Thanks for laying it out.  🚢

PR #16, opened 2022-10-13
feat: adding list API

pgautier404:
Adds push front, push back, and fetch operations.

### DIFF
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -2,10 +2,24 @@
 namespace Momento\Cache\CacheOperationTypes;
 
 use Cache_client\_GetResponse;
+use Cache_client\_ListFetchResponse;
 use Control_client\_ListCachesResponse;
 use Cache_client\_SetResponse;
 use Momento\Cache\Errors\MomentoErrorCode;
 use Momento\Cache\Errors\SdkError;
+
+class CacheInfo
+{
+    private string $name;
+
+    public function __construct($grpcListedCache) {
+        $this->name = $grpcListedCache->getCacheName();
+    }
+
+    public function name() : string {
+        return $this->name;
+    }
+}
 
 trait ErrorBody {
     private SdkError $innerException;
@@ -188,19 +202,6 @@ class ListCachesResponseError extends ListCachesResponse
     use ErrorBody;
 }
 
-class CacheInfo
-{
-    private string $name;
-
-    public function __construct($grpcListedCache) {
-        $this->name = $grpcListedCache->getCacheName();
-    }
-
-    public function name() : string {
-        return $this->name;
-    }
-}
-
 abstract class CacheSetResponse extends ResponseBase
 {
     public function asSuccess(): CacheSetResponseSuccess|null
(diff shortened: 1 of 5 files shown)


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -2,11 +2,25 @@
 namespace Momento\Cache\CacheOperationTypes;
 
 use Cache_client\_GetResponse;
+use Cache_client\_ListFetchResponse;
```
malandis, line 5: Is the generated code not in a separate namespace like `Momento_protos`?

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -322,3 +323,121 @@ class CacheDeleteResponseError extends CacheDeleteResponse
...
+    private array $values = [];
+
+    public function __construct(_ListFetchResponse $response)
+    {
+        parent::__construct();
+        if ($response->getFound())
+        {
+            foreach ($response->getFound()->getValues() as $value)
+            {
+                $this->values[] = $value;
```
cprice404, line 369: is this how you append to a list in PHP?  🤯 seems weird to be doing this in a loop

src/Cache/_ScsDataClient.php
```diff
@@ -38,7 +52,6 @@ class _ScsDataClient
     private static int $TIMEOUT_MULTIPLIER = 1000000;
```
malandis, line 52: What does this do? :thinking:

src/Cache/_ScsDataClient.php
```diff
@@ -120,9 +132,90 @@ public function delete(string $cacheName, string $key) : CacheDeleteResponse
...
+        try {
+            validateCacheName($cacheName);
+            validateListName($listName);
+            $itemTtlSeconds = $ttlSeconds ? $ttlSeconds : $this->defaultTtlSeconds;
+            validateTtl($itemTtlSeconds);
+            $listPushFrontRequest = new _ListPushFrontRequest();
+            $listPushFrontRequest->setListName($listName);
+            $listPushFrontRequest->setValue($value);
+            $listPushFrontRequest->setRefreshTtl($refreshTtl);
+            $listPushFrontRequest->setTtlMilliseconds($itemTtlSeconds * 1000);
```
malandis, line 176: I recommend pulling the seconds -> ms conversion to a separate function

tests/Cache/CacheClientTest.php
```diff
@@ -372,4 +372,72 @@ public function testDelete() {
...
+        $this->assertNotEmpty($values);
+        $this->assertCount(1, $values);
+        $this->assertContains($value, $values);
+        $response = $this->client->listPushFront($this->TEST_CACHE_NAME, $listName, 'value', true, 6000);
+        $this->assertNotNull($response->asSuccess());
+        $response = $this->client->listFetch($this->TEST_CACHE_NAME, $listName);
+        $this->assertNotNull($response->asHit());
+        $values = $response->asHit()->values();
+        $this->assertNotEmpty($values);
+        $this->assertCount(2, $values);
```
cprice404, line 394: nit - worth adding one more assertion to ensure the expected values in the expected order?

cprice404 (commented):
looking good to me, just a couple minor questions

malandis (commented):
Just a couple questions. Looks good

src/Cache/_ScsDataClient.php
pgautier404: There's a comment right above this explaining it.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: No, it generates its own namespaces which are separate from ours. See `composer.json` for more details.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: Yes, this is the syntax to append to a list. `getValues()` returns protobuf repeated fields that (as far as I know) I have to iterate over like this to extract the actual string values out of. This is not equivalent to `$this->values = $response->getFound()->getValues();` which returns a `TypeError`.

tests/Cache/CacheClientTest.php
pgautier404: Sure, added in 63c78c5.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
cprice404: yeah the iteration part makes sense to me if this is how you append to a list in php.  that is just, kind of the craziest syntax i've ever seen for appending to a list.  🤯

tests/Cache/CacheClientTest.php
cprice404: fantastic, thanks

cprice404 (dismissed):
yes

src/Cache/_ScsDataClient.php
pgautier404: Good idea, that let me DRY out some other stuff too. Thanks for the suggestion. Fixed in 18c813c.

cprice404 (approved):

PR #18, opened 2022-10-14
feat: adding list API continued

pgautier404:
Adds the remaining list API operations: pop front, pop back, length, and erase. Also adds `__toString()` in the response base class and overrides the base implementation in the response subtypes that hold data.

### DIFF
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -3,6 +3,9 @@ namespace Momento\Cache\CacheOperationTypes;
 
 use Cache_client\_GetResponse;
 use Cache_client\_ListFetchResponse;
+use Cache_client\_ListLengthResponse;
+use Cache_client\_ListPopBackResponse;
+use Cache_client\_ListPopFrontResponse;
 use Control_client\_ListCachesResponse;
 use Cache_client\_SetResponse;
 use Momento\Cache\Errors\MomentoErrorCode;
@@ -48,6 +51,12 @@ trait ErrorBody {
     {
         return $this->message;
     }
+
+    public function __toString()
+    {
+        return parent::__toString() . ": " . $this->message;
+    }
+
 }
 
 abstract class ResponseBase
@@ -81,6 +90,11 @@ abstract class ResponseBase
     protected function isMiss() : bool
     {
         return get_class($this) == "{$this->baseType}Miss";
+    }
+
+    public function __toString()
+    {
+        return get_class($this);
     }
 }
 
@@ -195,6 +209,12 @@ class ListCachesResponseSuccess extends ListCachesResponse
     {
         return $this->nextToken;
     }
+
+    public function __toString()
+    {
+        $cacheNames = array_map(fn($i) => $i->name(), $this->caches);
+        return get_class($this) . ": " . join(', ', $cacheNames);
+    }
 }
 
 class ListCachesResponseError extends ListCachesResponse
@@ -241,6 +261,10 @@ class CacheSetResponseSuccess extends CacheSetResponse
         return $this->value;
     }
 
+    public function __toString()
+    {
+        return get_class($this) . ": key {$this->key} = {$this->value}";
+    }
 }
 
 class CacheSetResponseError extends CacheSetResponse
@@ -290,6 +314,10 @@ class CacheGetResponseHit extends CacheGetResponse
         return $this->value;
     }
 
+    public function __toString()
+    {
+        return parent::__toString() . ": {$this->value}";
+    }
 }
 
 class CacheGetResponseMiss extends CacheGetResponse { }
@@ -358,6 +386,7 @@ class CacheListFetchResponseHit extends CacheListFetchResponse
 {
 
     private array $values = [];
+    private int $count;
 
     public function __construct(_ListFetchResponse $response)
     {
@@ -368,6 +397,7 @@ class CacheListFetchResponseHit extends CacheListFetchResponse
             {
                 $this->values[] = $value;
             }
+            $this->count = count($this->values);
         }
     }
 
@@ -376,6 +406,10 @@ class CacheListFetchResponseHit extends CacheListFetchResponse
         return $this->values;
     }
 
+    public function __toString()
+    {
+        return parent::__toString() . ": {$this->count} items";
+    }
 }
 
 class CacheListFetchResponseMiss extends CacheListFetchResponse { }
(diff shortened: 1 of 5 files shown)


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -441,3 +475,209 @@ class CacheListPushBackResponseError extends CacheListPushBackResponse
...
+        return null;
+    }
+
+    public function asError() : CacheListPopFrontResponseError|null
+    {
+        if ($this->isError())
+        {
+            return $this;
+        }
+        return null;
```
malandis, line 504: I like these conveniences. For my benefit: did we align on returning `null` when calling an incompatible call vs throwing an exception?

src/Cache/_ScsDataClient.php
```diff
@@ -224,4 +249,119 @@ public function listPushBack(
         return new CacheListPushBackResponseSuccess();
     }
 
+    public function listPopFront(string $cacheName, string $listName) : CacheListPopFrontResponse
+    {
+        try {
+            validateCacheName($cacheName);
+            validateListName($listName);
```
poppoerika, line 256: Don't we want to catch `InvalidArgumentError` for those?

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -441,3 +475,209 @@ class CacheListPushBackResponseError extends CacheListPushBackResponse
...
+    public function __construct(_ListPopFrontResponse $response)
+    {
+        parent::__construct();
+        $this->value = $response->getFound()->getFront();
+    }
+
+    public function value() : string
+    {
+        return $this->value;
+    }
```
malandis, line 520: Are we only doing `string` values here as opposed to both `string` and `bytes`?

src/Cache/_ScsDataClient.php
```diff
@@ -224,4 +249,119 @@ public function listPushBack(
...
+            return new CacheListPopFrontResponseMiss();
+        }
+        return new CacheListPopFrontResponseHit($response);
+    }
+
+    public function listPopBack(string $cacheName, string $listName) : CacheListPopBackResponse
+    {
+        try {
+            validateCacheName($cacheName);
+            validateListName($listName);
```
poppoerika, line 279: Same comment for catching `InvalidArgumentError`.

src/Cache/_ScsDataClient.php
```diff
@@ -224,4 +249,119 @@ public function listPushBack(
...
+            return new CacheListPopBackResponseMiss();
+        }
+        return new CacheListPopBackResponseHit($response);
+    }
+
+    public function listRemoveValue(string $cacheName, string $listName, string $value) : CacheListRemoveValueResponse
+    {
+        try {
+            validateCacheName($cacheName);
+            validateListName($listName);
```
poppoerika, line 302: Same comment for catching `InvalidArgumentError`.

src/Cache/_ScsDataClient.php
```diff
@@ -224,4 +249,119 @@ public function listPushBack(
...
+            return new CacheListRemoveValueResponseError(new UnknownError($e->getMessage()));
+        }
+        return new CacheListRemoveValueResponseSuccess();
+    }
+
+    public function listLength(string $cacheName, string $listName) : CacheListLengthResponse
+    {
+        try {
+            validateCacheName($cacheName);
+            validateListName($listName);
```
poppoerika, line 322: Same comment for catching `InvalidArgumentError`.

src/Cache/_ScsDataClient.php
```diff
@@ -224,4 +249,119 @@ public function listPushBack(
...
+        }
+        return new CacheListLengthResponseSuccess($response);
+    }
+
+    public function listErase(string $cacheName, string $listName, ?int $beginIndex=null, ?int $count=null) : CacheListEraseResponse
+    {
+        try {
+            validateCacheName($cacheName);
+            validateListName($listName);
+            validateRange($beginIndex, $count);
```
poppoerika, line 342: Same comment for catching `InvalidArgumentError`.

src/Cache/_ScsDataClient.php
```diff
@@ -224,4 +249,119 @@ public function listPushBack(
...
+            $listEraseRequest = new _ListEraseRequest();
+            $listEraseRequest->setListName($listName);
+            if (!is_null($beginIndex) && !is_null($count)) {
```
poppoerika, line 345: Do we need this if we are already checking with `validateRange`?

src/Utilities/_DataValidation.php
```diff
@@ -59,3 +59,41 @@ function validateOperationTimeout(?int $operationTimeout=null)
         }
     }
 }
+
+if (!function_exists('validateTruncateSize'))
```
malandis, line 63: Interesting. Is this a PHP idiom?

malandis (commented):
One question and a couple small questions. Looks great!

src/Cache/_ScsDataClient.php
poppoerika: Oh I see. Those are optional, and if they're null, delete all the items.

src/Utilities/_DataValidation.php
```diff
@@ -59,3 +59,41 @@ function validateOperationTimeout(?int $operationTimeout=null)
...
+        if ($truncateSize <= 0)
+        {
+            throw new InvalidArgumentError("Truncate size must be greater than zero.");
+        }
+    }
+}
+
+if (!function_exists('validateRange'))
+{
+    function validateRange(?int $beginIndex, ?int $count)
```
poppoerika, line 80: Do we want to prevent users from passing `count` value larger than the actual size of the list??

src/Cache/_ScsDataClient.php
pgautier404: We already do! It's an `SdkError` subclass which is handled in the first `catch` below.

src/Cache/_ScsDataClient.php
pgautier404: Yes. This is not a validation of the data. If these values are both supplied we need to erase a range of items and if neither value is supplied we need to erase the entire list. The validation makes sure that one or the other of those is true.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: Yes, we did. This allows for pseudo-pattern matching and enforces the contract with the developer that once a request is made to the cache client all Momento errors will be returned and not thrown.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: Yes, for reasons we discussed today.

src/Utilities/_DataValidation.php
pgautier404: It's the pattern I found while looking into autoloaded functions. I assume the autoloader may try to load the file multiple times under some circumstances, and this check prevents a fatal error about functions already being defined.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -290,6 +314,10 @@ public function value() : string
         return $this->value;
     }
 
+    public function __toString()
```
poppoerika, line 317: Is this PHP convention? i.e. Are PHP developers expected to `echo cacheGetResponseHit object`?
Sorry I misunderstand how `__toString()` works, but I didn't see any method like this in .NET SDK.

src/Utilities/_DataValidation.php
pgautier404: Validating that would require us to make a call to get the length of the list, and we don't want to make a round trip to the cache server for client side validation in my opinion. A number larger than the size of the list just deletes the rest of the list from the offset without complaining, so that round trip wouldn't buy us anything anyway. If you strongly disagree (or anyone else reading this disagrees) let me know and we can talk about it. 

That's a good thing to add to the tests in case the behavior ever changes . . . I'll do that.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: These are the PHP convention equivalent to the `ToString()` methods in .NET. I added one to each of the non-incubating SDK response types in https://github.com/momentohq/client-sdk-dotnet/pull/252. Your comment does raise the issue that these are **not** implemented in our incubating response types . . . I'll add a ticket for that on the .NET repo.

src/Utilities/_DataValidation.php
poppoerika: As I wrote this question, I was thinking the same thing about calling another API to get length will add latency so that's probably not something we want to do.
But I didn't know the behavior you explained when a number larger than the size of the list is passed, and I think it makes sense.

tests/Cache/CacheClientTest.php
```diff
@@ -444,4 +474,298 @@ public function testListPushBackFetchHappyPath() {
...
+
+        $response = $this->client->listErase($this->TEST_CACHE_NAME, $listName, 0, 2);
+        $this->assertNotEmpty($response->asSuccess());
+
+        $response = $this->client->listFetch($this->TEST_CACHE_NAME, $listName);
+        $this->assertNotNull($response->asHit());
+        $this->assertEquals(array_slice($values, 2), $response->asHit()->values());
+    }
+
+    public function testListEraseRange_LargeCountValue()
```
poppoerika, line 743: Thank you for adding this test!

poppoerika (dismissed):
Looks good to me!

tests/Cache/CacheClientTest.php
```diff
@@ -444,4 +474,298 @@ public function testListPushBackFetchHappyPath() {
...
+            $response = $this->client->listPushBack($this->TEST_CACHE_NAME, $listName, $val, false);
+            $this->assertNotNull($response->asSuccess());
+        }
+
+        $response = $this->client->listLength($this->TEST_CACHE_NAME, $listName);
+        $this->assertNotNull($response->asSuccess());
+        $this->assertEquals(4, $response->asSuccess()->length());
+
+        $response = $this->client->listErase($this->TEST_CACHE_NAME, $listName);
+        $this->assertNotEmpty($response->asSuccess());
```
poppoerika, line 713: This is also `assertNotNull`??

malandis (approved):
:1st_place_medal:

Issue #17, opened 2022-10-14
Add Dictionary APIs 


poppoerika:
- [x] CacheDictionaryDelete
- [x] CacheDictionaryFetch
- [x] CacheDictionaryGetBatch
- [x] CacheDictionaryGet
- [x] CacheDictionaryIncrement (@pgautier404 is going to help implementing this)
- [x] CacheDictionaryRemoveField (@pgautier404 is going to help implementing this)
- [x] CacheDictionaryRemoveFields (@pgautier404 is going to help implementing this)
- [x] CacheDictionarySetBatch
- [x] CacheDictionarySet

PR #19, opened 2022-10-16
chore: add dictionary set, get, delete

poppoerika:
Added dictionary set, get, and delete APIs and tests to check those operations.

Ticket: #17

### DIFF
--- a/src/Cache/_ScsDataClient.php
+++ b/src/Cache/_ScsDataClient.php
@@ -6,7 +6,6 @@ use Cache_client\_DeleteRequest;
 use Cache_client\_DictionaryDeleteRequest;
 use Cache_client\_DictionaryFieldValuePair;
 use Cache_client\_DictionaryGetRequest;
-use Cache_client\_DictionaryGetResponse;
 use Cache_client\_DictionarySetRequest;
 use Cache_client\_GetRequest;
 use Cache_client\_ListFetchRequest;
@@ -19,6 +18,8 @@ use Grpc\UnaryCall;
 use Momento\Cache\CacheOperationTypes\CacheDeleteResponse;
 use Momento\Cache\CacheOperationTypes\CacheDeleteResponseError;
 use Momento\Cache\CacheOperationTypes\CacheDeleteResponseSuccess;
+use Momento\Cache\CacheOperationTypes\CacheDictionaryDeleteResponseError;
+use Momento\Cache\CacheOperationTypes\CacheDictionaryDeleteResponseSuccess;
 use Momento\Cache\CacheOperationTypes\CacheDictionaryGetResponse;
 use Momento\Cache\CacheOperationTypes\CacheDictionaryGetResponseError;
 use Momento\Cache\CacheOperationTypes\CacheDictionaryGetResponseHit;
@@ -249,7 +250,7 @@ class _ScsDataClient
             $ttlMillis = $this->ttlToMillis($ttlSeconds);
             $dictionarySetRequest = new _DictionarySetRequest();
             $dictionarySetRequest->setDictionaryName($dictionaryName);
-            $dictionarySetRequest->setItems(toSingletonFieldValuePair($field, $value));
+            $dictionarySetRequest->setItems([$this->toSingletonFieldValuePair($field, $value)]);
             $dictionarySetRequest->setRefreshTtl($refreshTtl);
             $dictionarySetRequest->setTtlMilliseconds($ttlMillis);
             $call = $this->grpcManager->client->DictionarySet($dictionarySetRequest, ["cache" => [$cacheName]], ["timeout" => $this->deadline_seconds * self::$TIMEOUT_MULTIPLIER]);
@@ -264,6 +265,14 @@ class _ScsDataClient
         return new CacheDictionarySetResponseSuccess();
     }
 
+    private function toSingletonFieldValuePair(string $field, string $value): _DictionaryFieldValuePair
+    {
+        $singletonPair = new _DictionaryFieldValuePair();
+        $singletonPair->setField($field);
+        $singletonPair->setValue($value);
+        return $singletonPair;
+    }
+
     public function dictionaryGet(string $cacheName, string $dictionaryName, string $field): CacheDictionaryGetResponse
     {
         try {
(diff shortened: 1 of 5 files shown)


src/Cache/_ScsDataClient.php
```diff
@@ -114,25 +114,43 @@ public function set(string $cacheName, string $key, string $value, int $ttlSecon
...
+    }
+
+    private function processCall(UnaryCall $call): mixed
+    {
+        [$response, $status] = $call->wait();
+        if ($status->code !== 0) {
+            throw _ErrorConverter::convert($status->code, $status->details, $call->getMetadata());
+        }
+        return $response;
+    }
```
pgautier404, line 144: Can you move these definitions up so they live between `__construct()` and `set()` please?

src/Cache/_ScsDataClient.php
```diff
@@ -354,14 +369,92 @@ public function listErase(string $cacheName, string $listName, ?int $beginIndex=
...
+        } catch (SdkError $e) {
+            return new CacheDictionaryGetResponseError($e);
+        } catch (Exception $e) {
+            return new CacheDictionaryGetResponseError(new UnknownError($e->getMessage()));
+        }
+        if ($dictionaryGetResponse->hasMissing()) {
+            return new CacheDictionaryGetResponseMiss();
+        }
+        if ($dictionaryGetResponse->getFound()->getItems()->count() == 0) {
+            return new CacheDictionaryGetResponseError(new Exception("_DictionaryGetResponseResponse contained no data but was found"));
```
pgautier404, line 434: The exception used to construct the Error response class needs to be an SdkError like the one above this on line 428. Please use `UnknownError` like you did above unless you think one of the other SdkError-derived classes works better.

tests/Cache/CacheClientTest.php
```diff
@@ -117,26 +116,40 @@ public function testCreateCacheEmptyName()
...
 
+    private function getBadAuthTokenClient(): SimpleCacheClient
+    {
+        $badEnvName = "_MOMENTO_BAD_AUTH_TOKEN";
+        putenv("{$badEnvName}={$this->BAD_AUTH_TOKEN}");
+        $authProvider = new EnvMomentoTokenProvider($badEnvName);
+        putenv($badEnvName);
+        return new SimpleCacheClient($authProvider, $this->DEFAULT_TTL_SECONDS);
+    }
+
```
pgautier404, line 148: Can you move this to the top of the class, please, either before `setUp()` or after `tearDown()`?

tests/Cache/CacheClientTest.php
```diff
@@ -77,13 +67,18 @@ public function testCreateSetGetDelete() {
         $this->assertNotNull($response->asSuccess());
     }
 
-    // Client initialization tests
-    public function testNegativeDefaultTtl() {
+    // Happy path test
```
pgautier404, line 70: Please move this comment up so it appears before the `testCreateSetGetDelete` test.

tests/Cache/CacheClientTest.php
```diff
@@ -92,22 +87,26 @@ public function testNonJwtTokens() {
...
 
-    // Create cache tests
-    public function testCreateCacheAlreadyExists() {
+    public function testCreateCacheAlreadyExists()
+    {
         $response = $this->client->createCache($this->TEST_CACHE_NAME);
         $this->assertNotNull($response->asAlreadyExists());
     }
 
+    // Create cache tests
```
pgautier404, line 108: This comment should come before the `testCreateCacheAlreadyExists` test please.

poppoerika:
@pgautier404 Apologize for removing the list tests you've added. 
I need to be careful when I merge and use PhpStorm interface to manage these merges 🙇‍♀️ 
I addressed your feedback in this push.

pgautier404 (dismissed):
LGTM!

src/Cache/_ScsDataClient.php
```diff
@@ -354,14 +369,92 @@ public function listErase(string $cacheName, string $listName, ?int $beginIndex=
...
+            validateDictionaryName($dictionaryName);
+            validateFieldName($field);
+            validateValueName($value);
+            $ttlMillis = $this->ttlToMillis($ttlSeconds);
+            $dictionarySetRequest = new _DictionarySetRequest();
+            $dictionarySetRequest->setDictionaryName($dictionaryName);
+            $dictionarySetRequest->setItems([$this->toSingletonFieldValuePair($field, $value)]);
+            $dictionarySetRequest->setRefreshTtl($refreshTtl);
+            $dictionarySetRequest->setTtlMilliseconds($ttlMillis);
+            $call = $this->grpcManager->client->DictionarySet($dictionarySetRequest, ["cache" => [$cacheName]], ["timeout" => $this->deadline_seconds * self::$TIMEOUT_MULTIPLIER]);
```
malandis, line 396: I recommend pre-computing this so we don't multiply this out in all the functions, ie in the constructor store `deadline_microseconds = $this->deadline_seconds * self::$TIMEOUT_MULTIPLIER` and use that here

tests/Cache/CacheClientTest.php
```diff
@@ -768,4 +805,129 @@ public function testListErase_MissingList()
...
+        $response = $this->client->dictionarySet($this->TEST_CACHE_NAME, $dictionaryName, $field, $value, true, 10);
+        $this->assertNotNull($response->asSuccess());
+        sleep(2);
+
+        $response = $this->client->dictionaryGet($this->TEST_CACHE_NAME, $dictionaryName, $field);
+        $this->assertNotNull($response->asHit());
+        $this->assertEquals($value, $response->asHit()->value());
+    }
+
+    public function testDictionaryDeleteThrowExceptionForEmptyCacheName()
```
malandis, line 903: function name is now out of date -- returns error instead of throws exception :)

malandis (commented):
Few things:
- Recommend storing the deadline in microseconds to avoid duplicated code
- Recommend splitting data structures tests into multiple files, one for each data structure
- Consistent test name format (no underscores)

src/Cache/_ScsDataClient.php
pgautier404: I'd rather avoid using `deadline_microseconds` because they're technically not microseconds. I think this is a great idea, but I'd like to think about it a bit . . . I'll make this change in an upcoming PR.

poppoerika:
@malandis Thank you for the feedback!
Regarding splitting the tests, we'll address that in a follow-up PR. (https://github.com/momentohq/client-sdk-php/issues/22)
I will update the test names though.

poppoerika:
@malandis We will address test names in a follow-up pr to get this pr merge first.

pgautier404 (approved):

PR #30, opened 2022-10-20
feat: dict increment and remove

pgautier404:


src/Cache/_ScsDataClient.php
```diff
@@ -401,6 +411,7 @@ public function dictionarySet(string $cacheName, string $dictionaryName, string
             validateFieldName($field);
             validateValueName($value);
             $ttlMillis = $this->ttlToMillis($ttlSeconds);
+            validateTtl($ttlMillis);
```
poppoerika, line 414: Thank you!

src/Cache/_ScsDataClient.php
```diff
@@ -542,4 +553,82 @@ public function dictionaryGetBatch(string $cacheName, string $dictionaryName, ar
...
+    }
+
+    public function dictionaryRemoveFields(string $cacheName, string $dictionaryName, array $fields): CacheDictionaryRemoveFieldsResponse
+    {
+        try {
+            validateCacheName($cacheName);
+            validateDictionaryName($dictionaryName);
+            foreach ($fields as $field) {
+                validateFieldName($field);
+            }
```
poppoerika, line 616: Do we also want to validate `$fields` array is not an empty array?

poppoerika (commented):
Thank you for updating all the conflicts...!!!

cprice404 (dismissed):
🙈 i'm good when @poppoerika is

src/Cache/_ScsDataClient.php
pgautier404: Yes, we do! Thanks, good catch. Fixed in 6ea1091.

poppoerika (approved):
🎉

PR #34, opened 2022-10-21
chore: convert deadline secs to expected gRPC value in constructor

pgautier404:


malandis (approved):
Thanks! Looks cleaner and easier to maintain

PR #35, opened 2022-10-21
chore: add github release workflow

pgautier404:


.github/workflows/on-push-to-release.yaml
```diff
@@ -0,0 +1,67 @@
...
+        uses: momentohq/standards-and-practices/github-actions/shared-build@gh-actions-v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.MOMENTO_MACHINE_USER_GITHUB_TOKEN }}
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.1'
+          extensions: grpc
+          tools: composer
```
malandis, line 60: Is the gRPC extension also pinned?

malandis (commented):
Nice!

.github/workflows/on-push-to-release.yaml
pgautier404: This is the binary extension (shared object) that we can't really pin per project. This is something that has already been compiled and installed on the system, not a PHP library we download into `vendor`, so we have no guarantee of the versions we'll run into in the wild. 

I did try to pin the version to the one I'm relatively sure I have, but it is causing the PHP setup step of the workflow to hang indefinitely. I don't want to spend a lot of time diagnosing that when the utility of pinning the extension is not clear to me. We can talk about it and maybe cut a ticket to research it if you'd like, but this setup has been working fine from a CI perspective.

malandis (approved):

PR #45, opened 2022-10-28
chore: add `setAdd` and `setFetch`

poppoerika:
We need to add `setAdd` and `setFetch` to implement TaggedCache in Laravel. (`setFetch` is more to test `setAdd` is correctly implemented)


cprice404 (approved):
🙈

PR #46, opened 2022-11-01
chore: add Dockerfile to run tests

### LINKED ISSUE #27: create docker image for running php tests / examples
installing prerequisites on mac is proving to be an unreasonable amount of work for devs who just need to jump in and run tests etc.

poppoerika:
Also added instructions to build an image and run the tests in a container in both regular README and Japanese version.

Closes #27

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,15 +7,7 @@ RUN cd /usr/local/bin && mv composer.phar composer
 RUN pecl install grpc
 RUN docker-php-ext-enable grpc
 
-COPY . /app
-
 WORKDIR /app
-
+COPY composer.json composer.lock ./
 RUN composer install
-
-#ARG token
-#ARG cache_name
-#ENV TEST_AUTH_TOKEN=$token
-#ENV TEST_CACHE_NAME="php-integration-test-cache"
-#ENV MOMENTO_AUTH_TOKEN=$token
-#ENV CACHE_NAME=$cache_name
+COPY . ./
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,9 +13,9 @@ WORKDIR /app
 
 RUN composer install
 
-ARG token
-ARG cache_name
-ENV TEST_AUTH_TOKEN=$token
-ENV TEST_CACHE_NAME="php-integration-test-cache"
-ENV MOMENTO_AUTH_TOKEN=$token
-ENV CACHE_NAME=$cache_name
+#ARG token
+#ARG cache_name
+#ENV TEST_AUTH_TOKEN=$token
+#ENV TEST_CACHE_NAME="php-integration-test-cache"
+#ENV MOMENTO_AUTH_TOKEN=$token
+#ENV CACHE_NAME=$cache_name
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM php:8.1.12-cli
+
+RUN apt-get update && apt-get install -y -q git rake ruby-ronn zlib1g-dev && apt-get clean
+
+RUN cd /usr/local/bin && curl -sS https://getcomposer.org/installer | php
+RUN cd /usr/local/bin && mv composer.phar composer
+RUN pecl install grpc
+RUN docker-php-ext-enable grpc
+
+COPY . /app
+
+WORKDIR /app
+
+RUN composer install
+
+ARG token
+ENV TEST_AUTH_TOKEN=$token
+ENV TEST_CACHE_NAME="php-integration-test-cache"
+
+CMD ["php", "vendor/phpunit/phpunit/phpunit", "--configuration ", "phpunit.xml"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,5 @@ RUN pecl install grpc
 RUN docker-php-ext-enable grpc
 
 WORKDIR /app
-COPY composer.json composer.lock ./
-RUN composer install
 COPY . ./
+RUN composer install
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,8 @@ WORKDIR /app
 RUN composer install
 
 ARG token
+ARG cache_name
 ENV TEST_AUTH_TOKEN=$token
 ENV TEST_CACHE_NAME="php-integration-test-cache"
-
-CMD ["php", "vendor/phpunit/phpunit/phpunit", "--configuration ", "phpunit.xml"]
+ENV MOMENTO_AUTH_TOKEN=$token
+ENV CACHE_NAME=$cache_name
--- a/README.template.md
+++ b/README.template.md
@@ -59,4 +59,11 @@ Coming soon!
 
 Coming soon!
 
+### Running Tests
+
+```bash
+docker build --tag php-test --build-arg token=<YOUR_AUTH_TOKEN> .
+docker run -d php-test
+```
+
 {{ ossFooter }}
--- a/README.template.md
+++ b/README.template.md
@@ -59,12 +59,4 @@ Coming soon!
 
 Coming soon!
 
-### Running Tests
-
-```bash
-export DOCKER_COMMAND="php vendor/phpunit/phpunit/phpunit --configuration phpunit.xml"
-docker build --tag php-test --build-arg token=<YOUR_AUTH_TOKEN> .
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+WORKDIR /app
+
+RUN composer install
+
+ARG token
+ENV TEST_AUTH_TOKEN=$token
+ENV TEST_CACHE_NAME="php-integration-test-cache"
+
+CMD ["php", "vendor/phpunit/phpunit/phpunit", "--configuration ", "phpunit.xml"]
```
cprice404, line 20: can we make this a little more general purpose?  e.g. it could also be used for running examples.  I think we could just remove this CMD and then users can pass it at the end of their "docker run" command?

poppoerika: That makes sense! Will update Dockerfile and README.

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM php:8.1.12-cli
+
+RUN apt-get update && apt-get install -y -q git rake ruby-ronn zlib1g-dev && apt-get clean
+
+RUN cd /usr/local/bin && curl -sS https://getcomposer.org/installer | php
+RUN cd /usr/local/bin && mv composer.phar composer
+RUN pecl install grpc
+RUN docker-php-ext-enable grpc
```
cprice404, line 8: Everything above this line seems great and is working very well for me.

I'd like to see if we can keep this Dockerfile generic, though, so I have some thoughts on the remaining lines.

I think that in the not-too-distant future (not as part of this ticket or PR), I'd like to publish this image to a public ECR repo, something like `momento-php` or `momento-php-dev` or something.  Then we could re-use it across multiple projects, and devs could pull it without having to do the whole build (which is pretty expensive for the grpc part).

So with that in mind I think we might want to see if we can get rid of the remaining lines after this.  I will add some additional comments below.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+COPY . /app
+
+WORKDIR /app
```
cprice404, line 12: If we copy the app dir into the docker image then we have to re-build it every time we change the app, and also it makes this docker image very tightly coupled to this repo.

Instead I think we can use `-v` flags when we call `docker run`, and that will allow us to mount the appropriate directories into the docker container.  So I think we can get rid of these lines.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN composer install
```
cprice404, line 14: This part is obviously important and I can imagine that maybe it is expensive?

Do you know where composer installs files to?  Are they installed somewhere underneath the current working directory?  Or in a centralized cache that can be re-used by multiple PHP apps on the host?

I'm wondering if we can just have a generic composer.json file that we burn into the image, which would install all of our normal deps and maybe laravel too, and use `composer install` on that file in order to bake all of those deps into the generic image.  And then they could be re-used for various different apps that we might want to test with.  I'm not sure if that's possible or not.  Do you know?

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+ARG token
+ARG cache_name
+ENV TEST_AUTH_TOKEN=$token
+ENV TEST_CACHE_NAME="php-integration-test-cache"
+ENV MOMENTO_AUTH_TOKEN=$token
+ENV CACHE_NAME=$cache_name
```
cprice404, line 21: Again in the spirit of trying to make the image as re-usable as possible, and avoid the need to re-build it, I would get rid of all of these.  They can be passed in as env vars via `-e` flag to `docker run`.

README.template.md
```diff
@@ -59,4 +59,12 @@ Coming soon!
 
 Coming soon!
 
+### Running Tests
```
cprice404, line 62: These instructions are relevant to Momento devs but probably not to most OSS users.  So we should create a `CONTRIBUTING.md` file and put them in there instead of in the main README.

Take a look at the other SDK repos and see if we have an example of a `CONTRIBUTING.md` anywhere yet.  Ultimately we will want to have one in every SDK repo and we will want to make them somewhat consistent, but I'm not sure if we've really established a pattern for that yet.  If not, you can just get us started :)

I would recommend having two different sections in the CONTRIBUTING.md: one for how to build the docker image, and one for how to run things.  Our goal will be for devs to only need to build the docker image once (or very rarely), and then they should only need to do variations on `docker run` once they have the image.

README.template.md
```diff
@@ -59,4 +59,12 @@ Coming soon!
...
+
+```bash
+export DOCKER_COMMAND="php vendor/phpunit/phpunit/phpunit --configuration phpunit.xml"
+docker build --tag php-test --build-arg token=<YOUR_AUTH_TOKEN> .
+docker run -d -it php-test bash -c $DOCKER_COMMAND
```
cprice404, line 67: This didn't work for me; to get it to work I ran this command:

```
docker run -it php-test php vendor/phpunit/phpunit/phpunit --configuration phpunit.xml
```

When you follow my advice above about making the build more generic, you are going to end up with some additional args that are required for the docker run command; e.g. some `-e`s to set env vars, and probably a `-v` to mount the sdk code.  At that point the `docker run` command will be fairly long so it will probably be a good idea to have some bash scripts that people can use.  e.g. `dev-docker-build.sh`, and `dev-run-integration-tests.sh` or something like that.

src/Cache/_ScsDataClient.php
```diff
@@ -521,7 +520,6 @@ public function dictionarySetBatch(string $cacheName, string $dictionaryName, ar
             validateCacheName($cacheName);
             validateDictionaryName($dictionaryName);
             validateItems($items);
-            validateFieldsKeys($items);
```
cprice404, line 524: just curious what the deal with this one was

Dockerfile
poppoerika: I don't think baking cache into a image is not possible but we can cache composer packages for the first build and subsequent builds can export data from that cache.
I think we can use [this](https://docs.docker.com/build/building/cache/backends/).
I'd assume we will have build/push actions on GitHub action so maybe we're interested in using [this](https://docs.docker.com/build/building/cache/backends/gha/)?
Also, we can make it so that only new packages are installed in the subsequent builds by copying `composer.json` and `composer.lock` first and then run `composer install`.

Please let me know what you think!

cprice404: Don't worry about the automated build/push for now, we can tackle that a little later.  but thanks for thinking about it.

As for the caching of composer stuff via that first link you sent, it might be good for us to have a zoom to brainstorm on the ideas there.

The idea of copying in the composer files and then doing composer install is definitely along the lines of what I was thinking.

Dockerfile
poppoerika: I think with Dockerfile it might be difficult to achieve this.
What I have tried during docker build:
- Copy only `composer.json` and `composer.lock` 
- Run `composer install` 
- Results in creation of `vender` which includes all the packages' files

And at `docker run`, set a volume to mount client-sdk-php to a Docker container.
But what happens is that this volume overwrites (well dah) `vendor` and now there are no packages installed.

It looks like we could specify which directories/files to mount using `docker-compose.yml`. But I don't think that achieve what we're trying to do, which re-using this image for other purposes.

Dockerfile
poppoerika: I will do this.

src/Cache/_ScsDataClient.php
poppoerika: Hmm sorry I thought I added it back after trying to figure out why it failed. I'm adding this back.

Dockerfile
cprice404: okay, I will pull this down and look into it when i get a chance.  It might be tomorrow.

I suspect we can just mount the vendor directory to keep the image generic but I need to double check to make sure it doesn't contain any native extensions or anything like that.

poppoerika:
@cprice404 I think class mapping in `composer.json` is causing an issue to copy `composer.json` and `composer.lock` first and run `composer install` to build a Docker image.
I know this is one way to optimize Docker build but I'm leaving it as it is originally (copying the entire files and then run `composer install`).

cprice404 (approved):
containers

this looks good for now!  i will play with the composer part later when i get a chance.  thanks!

PR #55, opened 2022-11-03
chore: add PSR-16 library

pgautier404:


README-PSR16.md
```diff
@@ -0,0 +1,33 @@
...
+full [PSR-16 library example](https://github.com/momentohq/client-sdk-php/blob/main/examples/psr16-example.php) in the
+[examples directory](https://github.com/momentohq/client-sdk-php/blob/main/examples) of this repository!
+
+### Implementation Notes
+
+- Please note that this library is under active development and may be subject to backward-incompatible changes. In
+  addition, the `clear()` function defined in the PSR-16 specification is currently unimplemented in this client
+  library (but is coming soon, so stay tuned!).
+- The `getMultiple()`, `setMultiple()`, and `deleteMultiple()` functionality is currently implemented client-side and
+  may exhibit slower performance that expected. These methods will be replaced with calls to server-side implementations
```
poppoerika, line 32: `slower performance than`?

README.template.md
```diff
@@ -51,9 +51,53 @@ Here is an example to get you started:
...
+```
+
+Using this approach, you get a type-safe `hitResponse` object in the case of a cache hit. But if the cache read
+results in a Miss or an error, you'll also get a type-safe object that you can use to get more info about what happened.
+
+In cases where you get an error response, `Error` types will always include an `ErrorCode` that you can use to check
+the error type:
+
+```php
+$getResponse = $client->get(CACHE_NAME, KEY);
```
poppoerika, line 89: nit: `$getResponse = $client->get($CACHE_NAME, $KEY);`?

examples/psr16-example.php
```diff
@@ -0,0 +1,53 @@
+<?php
+declare(strict_types=1);
+
+require "vendor/autoload.php";
```
poppoerika, line 4: What does this do?

README-PSR16.md
```diff
@@ -0,0 +1,33 @@
+# Momento PSR-16 Client Library
+
+The Momento PSR-16 client library, `Psr16SImpleCache`, implements the [PHP PSR-16 common interface for caching
```
cprice404, line 3: I think your capital `I` should be lower case here?  That really through me off for a sec, I was like "what's an `ImpleCache`?"

README-PSR16.md
```diff
@@ -0,0 +1,33 @@
...
+### Examples
+
+Check out the
+full [PSR-16 library example](https://github.com/momentohq/client-sdk-php/blob/main/examples/psr16-example.php) in the
+[examples directory](https://github.com/momentohq/client-sdk-php/blob/main/examples) of this repository!
+
+### Implementation Notes
+
+- Please note that this library is under active development and may be subject to backward-incompatible changes. In
+  addition, the `clear()` function defined in the PSR-16 specification is currently unimplemented in this client
```
cprice404, line 29: does `clear` through?  Should we explicitly state that?  Would it be worth separating that as its own bullet point?

README-PSR16.md
```diff
@@ -0,0 +1,33 @@
...
+  library (but is coming soon, so stay tuned!).
+- The `getMultiple()`, `setMultiple()`, and `deleteMultiple()` functionality is currently implemented client-side and
+  may exhibit slower performance that expected. These methods will be replaced with calls to server-side implementations
+  when they are available.
```
cprice404, line 33: maybe just say "in the near future"?

README.template.md
```diff
@@ -51,9 +51,53 @@ Here is an example to get you started:
 {{ usageExampleCode }}
 ```
 
+In addition to the `SimpleCacheClient` client library shown in the example above, a PHP PSR-16 "SimpleCache"
+implementation is also included in the SDK. See the [PSR-16 client README](README-PSR16.md) for more details.
```
cprice404, line 55: Do you think it might be worth pushing this up into the `Examples` section to make it more visible?

README.template.md
```diff
@@ -51,9 +51,53 @@ Here is an example to get you started:
...
 
-Coming soon!
+Errors that occur in calls to `SimpleCacheClient` methods are surfaced to developers as part of the return values of
+the calls, as opposed to by throwing exceptions. This makes them more visible, and allows your IDE to be more
+helpful in ensuring that you've handled the ones you care about. (For more on our philosophy about this, see our
+blog post on why [Exceptions are bugs](https://www.gomomento.com/blog/exceptions-are-bugs). And send us any
+feedback you have!)
+
+The preferred way of interpreting the return values from SimpleCacheClient methods is
+using [Pattern matching](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching).
```
cprice404, line 66: you probably don't want the .NET link here.

README.template.md
```diff
@@ -51,9 +51,53 @@ Here is an example to get you started:
...
+{
+    if ($errorResponse->errorCode() == MomentoErrorCode.TIMEOUT_ERROR) {
+       // this would represent a client-side timeout, and you could fall back to your original data source
+    }
+}
+```
+
+Note that, outside of SimpleCacheClient responses, exceptions can occur and should be handled as usual. For example,
+trying to instantiate a SimpleCacheClient with an invalid authentication token will result in an
+IllegalArgumentException being thrown.
```
cprice404, line 100: thanks for adding this!

composer.json
```diff
@@ -27,7 +27,9 @@
     "ext-grpc": "*",
     "firebase/php-jwt": "^6.3",
     "google/protobuf": "3.21.5",
-    "grpc/grpc": "1.42.0"
+    "grpc/grpc": "1.42.0",
+    "monolog/monolog": "^3.2",
+    "psr/simple-cache": "^3.0"
```
cprice404, line 32: so this means that our main SDK now has these new deps, correct?

Just food for thought - whether this might cause transitive dependency conflicts for existing apps.  We generally want to try to keep the dep list as minimal as possible.

It might be worth packaging the PSR implementation separately for this reason, but I have mixed feelings about that because I want it to be easily discoverable.

No change requested for now but just something to think about.  If we keep it like this for 1.0 then I think it would technically require a 2.0 release to separate them out.

src/Utilities/LoggingHelper.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Momento\Utilities;
+
+use Monolog\Formatter\LineFormatter;
+use Monolog\Handler\NullHandler;
+use Monolog\Handler\StreamHandler;
+use Monolog\Level;
+use Monolog\Logger;
+
+class LoggingHelper
```
cprice404, line 12: not a blocker for this PR, but just want to make sure that we are thinking about we can ensure that the user's pre-existing logger/config will be respected if they have one.  If that's possible.  I'm not sure what the PHP logging ecosystem looks like.

tests/Cache/Psr16ClientTest.php
```diff
@@ -0,0 +1,311 @@
...
+        $this->assertNotNull($this->client->get($key1));
+        $this->assertNotNull($this->client->get($key2));
+
+        sleep(4);
+
+        $this->assertNull($this->client->get($key1));
+        $this->assertNull($this->client->get($key2));
+    }
+
+}
```
cprice404, line 311: thanks for adding these tests!

cprice404 (commented):
docs nits and a few food-for-thought questions but this is looking great!

examples/psr16-example.php
pgautier404: The `declare(strict_types=1)` makes PHP act like a strictly typed language and throw an exception when you try to use, for example, a string where an int is required. Without that, PHP will try to cast the value to the required type instead. The `require` loads the autogenerated code the PHP autoloader needs.

README.template.md
pgautier404: Good idea, I'm moving it up.

README.template.md
pgautier404: Haha, wow. Yes, you are correct.

src/Utilities/LoggingHelper.php
pgautier404: Good to call out, thanks. Monolog is one of a large number of logging packages for PHP. I chose it partly because it is also integrated with or (supported via plugin) by all of the main PHP app frameworks. 

There is a logging PSR, PSR-3, that Monolog supports and that may allow me to generalize what we can accept from users. I'll do some looking.

cprice404 (approved):
compliance

PR #58, opened 2022-11-08
fix: fix push to release branch action

pgautier404:


poppoerika (approved):
I should update laravel-cache and laravel-example repos' release workflow too.

PR #60, opened 2022-11-11
feat: add configuration class

pgautier404:
These changes bring the PHP SDK into closer alignment with standards and practices regarding configuration and logging. It removes the `Monolog` PHP logger package dependency from the SDK and adds a null logger and a (very very) minimal logger for users who don't want to provide their own. The SDK now accepts any logging object from the user that implements the PSR-3 PHP logging standard. It also updates the examples to show both Monolog logging and examples of the included loggers. 

This PR also replaces the previous basic metadata implementation with actual interceptor classes and corrects the previous behavior of sending the user agent information to the server on each request.

Finally, this PR configures `composer.json` to calculate dependencies using version 8.0 of PHP regardless of which version of PHP is used to execute `composer update`.


src/Cache/Interceptors/AgentInterceptor.php
```diff
@@ -0,0 +1,23 @@
...
+class AgentInterceptor extends Interceptor
+{
+
+    private const AGENT = "php:0.3.0";
+    private bool $isFirstRequest = true;
+
+    public function interceptUnaryUnary($method, $argument, $deserialize, $continuation, array $metadata = [], array $options = [])
+    {
+        if ($this->isFirstRequest) {
+            $metadata["agent"] = [self::AGENT];
```
schwern, line 17: If a user sets their own agent, this will silently overwrite it. Do we want that?

pgautier404: The caller can't specify an agent. This is an internal value for metrics purposes.

src/Cache/Psr16CacheClient.php
```diff
@@ -24,14 +25,20 @@ class Psr16CacheClient implements CacheInterface
...
+    public function __construct(
+        IConfiguration      $configuration,
+        ICredentialProvider $authProvider,
+        ?int                $defaultTtlSeconds,
+        ?bool               $throwExceptions = null
+    )
     {
         $ttlSeconds = $defaultTtlSeconds ?? self::DEFAULT_TTL_SECONDS;
-        $this->momento = new SimpleCacheClient($authProvider, $ttlSeconds);
+        $this->momento = new SimpleCacheClient($configuration, $authProvider, $ttlSeconds);
```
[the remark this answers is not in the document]
pgautier404: If we were at a 1.0 already and making this change I'd be more concerned about it, but at this stage this isn't going to pull the rug out from under anyone. I think the docs, examples, and IDE hints should help keep people straight.

composer.json
```diff
@@ -28,14 +28,17 @@
     "firebase/php-jwt": "^6.3",
     "google/protobuf": "3.21.5",
     "grpc/grpc": "1.42.0",
-    "monolog/monolog": "^2.5",
+    "psr/log": "^2.0",
```
cprice404, line 31: cool

examples/dictionary-example.php
```diff
@@ -1,28 +1,49 @@
...
 $FIELD = "MyField";
 $VALUE = "MyValue";
 $ITEM_DEFAULT_TTL_SECONDS = 60;
 
-function printBanner(string $message): void
+// Setup
+$authProvider = new EnvMomentoTokenProvider("MOMENTO_AUTH_TOKEN");
+
+// Use your favorite PSR-3 logger.
+$logger = new Logger("example");
```
cprice404, line 24: I like what you are doing here but just one question - does PSR-3 specify an interface for a logger *factory* too?  Or just for a logger?

The reason I ask is that in most languages the logging frameworks will allow/encourage you to instantiate a logger scoped to a specific class or module, so that each log message can clearly identify where it came from.  With the approach you have here we sort of assume a fairly global/monolithic logger that won't clearly distinguish which log messages come from us vs come from other sources.  Am I reading that right?

It's also entirely possible that that granular logging concept isn't common in PHP and that this is indeed the idiomatic / best-practice way to do this, but I wanted to ask.

cprice404 (commented):
looking great, thanks for doing this.  just one question about the php logging api.

examples/dictionary-example.php
pgautier404: There isn't a concept of factories in [PSR-3](https://www.php-fig.org/psr/psr-3/). In the [Monolog](https://github.com/Seldaek/monolog) example above, the string "example" in the constructor is the logging channel name, and that channel name can be included in log messages if desired. Monolog can instantiate multiple loggers with multiple channel names and point each one to different output stream or service with custom formats. It's very configurable, kind of to the point where I'm not sure it's possible to expose a single factory capable of instantiating objects from any arbitrary PSR-3 implementation.

That said, I'll look into some of the other PSR-3 implementations and see if there is any commonality I can use to provide some kind of factory.

pgautier404: Yeah, here's a taste of the diversity of logging approaches: https://www.loggly.com/ultimate-guide/php-logging-libraries/. I don't think the factory approach is feasible in this ecosystem unless we choose/bless a specific logging solution and add a factory for it, which defeats the purpose of generalizing to PSR-3 in the first place.

cprice404: okay let's merge this.  i did have one other thought i'd like to chat with you about for 5 mins but it doesn't block this PR.

cprice404 (approved):

PR #63, opened 2022-11-11
chore: prefix all value accessors with `value<type>`

### LINKED ISSUE #47: prefix all value accessors with `.Value`

### LINKED ISSUE #48: Make all value accessors be properties

poppoerika:
Updated all value accessors with `value<type>`. 
For example, `valueString` and `valueInt`. 

As commented [here](https://github.com/momentohq/client-sdk-php/issues/48#issuecomment-1311988727), I think we can close #48. (I'm mentioning it here since I was going to address this issue in this PR along with #47).

Closes #47

### DIFF
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -1012,7 +1012,7 @@ class CacheDictionaryGetBatchResponseSuccess extends CacheDictionaryGetBatchResp
         $ret = [];
         foreach ($this->responsesList as $response) {
             if ($response->asHit()) {
-                $ret[] = $response->asHit()->value();
+                $ret[] = $response->asHit()->valueString();
             }
             if ($response->asMiss()) {
                 $ret[] = null;
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -430,7 +430,7 @@ class CacheListFetchResponseHit extends CacheListFetchResponse
         }
     }
 
-    public function valuesString(): array
+    public function valuesArray(): array
     {
         return $this->values;
     }
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -284,7 +284,7 @@ class CacheSetResponseSuccess extends CacheSetResponse
         return $this->key;
     }
 
-    public function value(): string
+    public function valueString(): string
     {
         return $this->value;
     }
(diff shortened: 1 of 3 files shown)


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -1007,12 +1007,12 @@ public function __construct(_DictionaryGetResponse $responses = null, ?int $numR
         }
     }
 
-    public function values(): array
+    public function valuesArray(): array
```
cprice404, line 1010: oh right, a dictionary in php is actually an array?

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -430,7 +430,7 @@ public function __construct(_ListFetchResponse $response)
         }
     }
 
-    public function values(): array
+    public function valuesString(): array
```
cprice404, line 433: this isn't returning a string though?

cprice404 (commented):
the ones i see here lgtm other than the one String one that i called out that i think should be an array.  other than that I'm good when @pgautier404 is, if this is the full list.

i guess one thing i didn't fully consider on this one though is that in .NET we needed to include the types in the names because of the "byte" vs "string" flavors.  In PHP since we don't have that "byte" ambiguity we may not need to be quite so explicit on these.  I should have thought of this sooner and asked for @pgautier404 's input.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
poppoerika: Yeah, PHP's array is an array and a dictionary.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -1063,12 +1063,12 @@ public function __construct(_DictionaryIncrementResponse $response)
...
     }
 
-    public function value(): int
+    public function valueInt(): int
     {
         return $this->value;
     }
 
-    public function string(): string
+    public function valueString(): string
```
pgautier404, line 1071: Would you mind just removing this function please? I added it originally, apparently by mistake.

pgautier404:
> the ones i see here lgtm other than the one String one that i called out that i think should be an array. other than that I'm good when @pgautier404 is, if this is the full list.
> 
> i guess one thing i didn't fully consider on this one though is that in .NET we needed to include the types in the names because of the "byte" vs "string" flavors. In PHP since we don't have that "byte" ambiguity we may not need to be quite so explicit on these. I should have thought of this sooner and asked for @pgautier404 's input.

Adding the data type does seem a little verbose given we're generally only ever providing a single type. I'm fine with the verbosity, though, personally. Having the data type in the name does provide an extra hint/sanity check for users.

pgautier404 (approved):
Thank you!

PR #64, opened 2022-11-15
chore: update dictionary/set method names

poppoerika:
Updated dictionary/set methods names in `SimpleCacheClient` as directed in #52.
Also, updated the names in the `CacheClientTest` accordingly.

Closes #52

PR #65, opened 2022-11-15
chore:  add `fieldValueDictionary` accessor to `dictionaryGetBatch` 

### LINKED ISSUE #53: `DictionaryGetBatch` should return a dictionary

poppoerika:
Added `fieldValueDictionary` accessor to `dictionaryGetBatch` to return an array of field/value pairs.
Also, updated `dictionary` accessor to `fieldValueDictionary` for `dictionaryFetch` to be consistent with naming.
Corresponding tests are also updated.

Closes #53

### DIFF
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -982,8 +982,8 @@ abstract class CacheDictionaryGetFieldsResponse extends ResponseBase
 
 class CacheDictionaryGetFieldsResponseSuccess extends CacheDictionaryGetFieldsResponse
 {
-    private array $responsesList = [];
-    private array $fieldValueDictionary = [];
+    private array $responses = [];
+    private array $valuesDictionary = [];
 
     public function __construct(_DictionaryGetResponse $responses = null, ?int $numRequested = null, ?array $fields = null)
     {
@@ -992,46 +992,37 @@ class CacheDictionaryGetFieldsResponseSuccess extends CacheDictionaryGetFieldsRe
             parent::__construct();
             foreach ($responses->getFound()->getItems() as $response) {
                 if ($response->getResult() == ECacheResult::Hit) {
-                    $this->responsesList[] = new CacheDictionaryGetFieldResponseHit(null, $response->getCacheBody());
-                    $this->fieldValueDictionary[$fields[$counter]] = $response->getCacheBody();
+                    $this->responses[] = new CacheDictionaryGetFieldResponseHit(null, $response->getCacheBody());
+                    $this->valuesDictionary[$fields[$counter]] = $response->getCacheBody();
                     $counter++;
                 }
                 if ($response->getResult() == ECacheResult::Miss) {
-                    $this->responsesList[] = new CacheDictionaryGetFieldResponseMiss();
+                    $this->responses[] = new CacheDictionaryGetFieldResponseMiss();
                 } else {
-                    $this->responsesList[] = new CacheDictionaryGetFieldResponseError(new UnknownError(strval($response->getResult())));
+                    $this->responses[] = new CacheDictionaryGetFieldResponseError(new UnknownError(strval($response->getResult())));
                 }
             }
         }
         if (is_null($responses) && !is_null($numRequested)) {
             foreach (range(0, $numRequested - 1) as $ignored) {
-                $this->responsesList[] = new CacheDictionaryGetFieldResponseMiss();
+                $this->responses[] = new CacheDictionaryGetFieldResponseMiss();
             }
         }
     }
 
-    public function valuesArray(): array
+    public function responses(): array
     {
-        $ret = [];
-        foreach ($this->responsesList as $response) {
-            if ($response->asHit()) {
-                $ret[] = $response->asHit()->valueString();
-            }
-            if ($response->asMiss()) {
-                $ret[] = null;
-            }
-        }
-        return $ret;
+        return $this->responses;
     }
 
-    public function fieldValueDictionary(): array
+    public function valuesDictionary(): array
     {
-        return $this->fieldValueDictionary;
+        return $this->valuesDictionary;
     }
 
     public function __toString()
     {
-        $numResponses = count($this->responsesList);
+        $numResponses = count($this->responses());
         return parent::__toString() . ": $numResponses responses";
     }
 }
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -995,8 +995,7 @@ class CacheDictionaryGetFieldsResponseSuccess extends CacheDictionaryGetFieldsRe
                     $this->responses[] = new CacheDictionaryGetFieldResponseHit(null, $response->getCacheBody());
                     $this->valuesDictionary[$fields[$counter]] = $response->getCacheBody();
                     $counter++;
-                }
-                if ($response->getResult() == ECacheResult::Miss) {
+                } else if ($response->getResult() == ECacheResult::Miss) {
                     $this->responses[] = new CacheDictionaryGetFieldResponseMiss();
                 } else {
                     $this->responses[] = new CacheDictionaryGetFieldResponseError(new UnknownError(strval($response->getResult())));
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -983,7 +983,7 @@ abstract class CacheDictionaryGetBatchResponse extends ResponseBase
 class CacheDictionaryGetBatchResponseSuccess extends CacheDictionaryGetBatchResponse
 {
     private array $responsesList = [];
-    private array $fieldsValuesList = [];
+    private array $fieldValueDictionary = [];
 
     public function __construct(_DictionaryGetResponse $responses = null, ?int $numRequested = null, ?array $fields = null)
     {
@@ -993,7 +993,7 @@ class CacheDictionaryGetBatchResponseSuccess extends CacheDictionaryGetBatchResp
             foreach ($responses->getFound()->getItems() as $response) {
                 if ($response->getResult() == ECacheResult::Hit) {
                     $this->responsesList[] = new CacheDictionaryGetResponseHit(null, $response->getCacheBody());
-                    $this->fieldsValuesList[$fields[$counter]] = $response->getCacheBody();
+                    $this->fieldValueDictionary[$fields[$counter]] = $response->getCacheBody();
                     $counter++;
                 }
                 if ($response->getResult() == ECacheResult::Miss) {
@@ -1024,9 +1024,9 @@ class CacheDictionaryGetBatchResponseSuccess extends CacheDictionaryGetBatchResp
         return $ret;
     }
 
-    public function fieldsValuesArray(): array
+    public function fieldValueDictionary(): array
     {
-        return $this->fieldsValuesList;
+        return $this->fieldValueDictionary;
     }
 
     public function __toString()
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -908,7 +908,7 @@ class CacheDictionaryFetchResponseHit extends CacheDictionaryFetchResponse
         parent::__construct();
         $items = $response->getFound()->getItems();
         foreach ($items as $item) {
-            $this->$valuesDictionary[$item->getField()] = $item->getValue();
+            $this->valuesDictionary[$item->getField()] = $item->getValue();
         }
     }
 
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -901,25 +901,25 @@ abstract class CacheDictionaryFetchResponse extends ResponseBase
 
 class CacheDictionaryFetchResponseHit extends CacheDictionaryFetchResponse
 {
-    private array $dictionary;
+    private array $fieldValueDictionary;
 
     public function __construct(_DictionaryFetchResponse $response)
     {
         parent::__construct();
         $items = $response->getFound()->getItems();
         foreach ($items as $item) {
-            $this->dictionary[$item->getField()] = $item->getValue();
+            $this->fieldValueDictionary[$item->getField()] = $item->getValue();
         }
     }
 
-    public function dictionary(): array
+    public function fieldValueDictionary(): array
     {
-        return $this->dictionary;
+        return $this->fieldValueDictionary;
     }
 
     public function __toString()
     {
-        $numItems = count($this->dictionary);
+        $numItems = count($this->fieldValueDictionary);
         return parent::__toString() . ": $numItems items";
     }
 }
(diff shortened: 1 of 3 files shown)


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -1020,6 +1024,11 @@ public function valuesArray(): array
         return $ret;
     }
 
+    public function fieldValueDictionary(): array
```
cprice404, line 1027: I think by our naming conventions this would start with `value`.  I think maybe it would be `valueDictionary` but let's get @pgautier404 to weigh in.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -1020,6 +1024,11 @@ public function valuesArray(): array
         return $ret;
```
cprice404, line 1024: also i think we might want to change the name of `valuesArray` to just `responses`?  Since it's not actually returning values directly, but rather, an array of objects that may have values in them?

src/Cache/_ScsDataClient.php
```diff
@@ -576,7 +576,7 @@ public function dictionaryGetBatch(string $cacheName, string $dictionaryName, ar
             return new CacheDictionaryGetBatchResponseError(new UnknownError($e->getMessage()));
         }
         if ($dictionaryGetBatchResponse->hasFound()) {
-            return new CacheDictionaryGetBatchResponseSuccess($dictionaryGetBatchResponse);
+            return new CacheDictionaryGetBatchResponseSuccess($dictionaryGetBatchResponse, fields: $fields);
```
cprice404, line 579: I think you have another PR up that may have conflicts with this one, yes?  Where this class gets renamed to `*DictionaryGetFields*`?  Just calling it out to make sure that we end up with the correct one before the final merge.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: Yes, everything should begin with `value`. I thought we were going to go with something like `valueArray()` because dictionaries are arrays in PHP, but I could go with either name as long as it's consistent.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: @cprice404 this actually is extracting the values from the responses and returning them.

src/Cache/_ScsDataClient.php
poppoerika: Yeah this is the PR for that: https://github.com/momentohq/client-sdk-php/pull/64

poppoerika:
@pgautier404 @cprice404 
We want to return two values from `dictionaryGetFields` (which is `dictionaryGetBatch` in this branch still)
1. an array containing `CacheDictionaryGetResponse`'s values; in case of HIT, it's a string value, and in vase of MISS, it's `null` (currently this is called `valuesArray`)
2. an array containing field/value pairs, which I named `fieldValueDictionary`.

For 1, I think we can keep the name as is. Or we can rename this to `responses` to align with our .NET SDK, although it doesn't follow the `value<type>` naming convention.
For 2, I wanted to hear your opinion. I chose `fieldValueDictionary` because I wanted to differentiate it from `valuesArray` since it contains field/value pairs compared to just values. How about `fieldStringValueStringArray` to represent a content inside? 
This suggestion will be changed depending on our decision for 1.

pgautier404:
> @pgautier404 @cprice404 We want to return two values from `dictionaryGetFields` (which is `dictionaryGetBatch` in this branch still)
> 
>     1. an array containing `CacheDictionaryGetResponse`'s values; in case of HIT, it's a string value, and in vase of MISS, it's `null` (currently this is called `valuesArray`)
> 
>     2. an array containing field/value pairs, which I named `fieldValueDictionary`.
> 
> 
> For 1, I think we can keep the name as is. Or we can rename this to `responses` to align with our .NET SDK, although it doesn't follow the `value<type>` naming convention. For 2, I wanted to hear your opinion. I chose `fieldValueDictionary` because I wanted to differentiate it from `valuesArray` since it contains field/value pairs compared to just values. How about `fieldStringValueStringArray` to represent a content inside? This suggestion will be changed depending on our decision for 1.

The .NET SDK is returning responses without extracting values from them in a `Responses` property. There are also methods for returning the extracted values of responses as a list of strings or a list of byte arrays. It doesn't return key/value pairs anywhere. We need to do the same. Like Chris suggested, the function returning the list of responses should just be called `responses`, and the list of extracted values should be called `valuesString()`.

@cprice404 @malandis can you sanity check me here please?

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: See https://github.com/momentohq/client-sdk-php/pull/65#issuecomment-1315780503

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: See https://github.com/momentohq/client-sdk-php/pull/65#issuecomment-1315780503

cprice404:
> > @pgautier404 @cprice404 We want to return two values from `dictionaryGetFields` (which is `dictionaryGetBatch` in this branch still)
> > ```
> > 1. an array containing `CacheDictionaryGetResponse`'s values; in case of HIT, it's a string value, and in vase of MISS, it's `null` (currently this is called `valuesArray`)
> > 
> > 2. an array containing field/value pairs, which I named `fieldValueDictionary`.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For 1, I think we can keep the name as is. Or we can rename this to `responses` to align with our .NET SDK, although it doesn't follow the `value<type>` naming convention. For 2, I wanted to hear your opinion. I chose `fieldValueDictionary` because I wanted to differentiate it from `valuesArray` since it contains field/value pairs compared to just values. How about `fieldStringValueStringArray` to represent a content inside? This suggestion will be changed depending on our decision for 1.
> 
> The .NET SDK is returning responses without extracting values from them in a `Responses` property. There are also methods for returning the extracted values of responses as a list of strings or a list of byte arrays. It doesn't return key/value pairs anywhere. We need to do the same. Like Chris suggested, the function returning the list of responses should just be called `responses`, and the list of extracted values should be called `valuesString()`.
> 
> @cprice404 @malandis can you sanity check me here please?

My intention for this API is that there will only be two accessors on this response type after we make the changes:
1. A values dictionary
2. An array containing the raw responses (values not extracted, so no `.value` prefix necessary)

In .NET we will be removing the methods/properties for "accessing the extracted values as a list of strings or as a list of byte arrays".

src/Cache/CacheOperationTypes/CacheOperationTypes.php
cprice404: an https://github.com/momentohq/client-sdk-php/pull/65#issuecomment-1315789863

cprice404:
> Should I use dictionary instead of valuesArray although dictionary is not a type in PHP?

It should start with `values` either way.  `valuesDictionary` makes more sense to me but it might not for folks more experienced with PHP.  @pgautier404 wdyt?

pgautier404:
> > Should I use dictionary instead of valuesArray although dictionary is not a type in PHP?
> 
> It should start with `values` either way. `valuesDictionary` makes more sense to me but it might not for folks more experienced with PHP. @pgautier404 wdyt?

I honestly couldn't speak for PHP devs other than myself, but I would personally prefer `valuesDictionary` and appreciate the hint about which of the polymorphic PHP `array` types I was supposed to be getting. When I type `$response->asHit()->v` I think the completion should help document how to handle the value type.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -982,47 +982,46 @@ public function asError(): CacheDictionaryGetFieldsResponseError|null
...
             parent::__construct();
             foreach ($responses->getFound()->getItems() as $response) {
                 if ($response->getResult() == ECacheResult::Hit) {
-                    $this->responsesList[] = new CacheDictionaryGetFieldResponseHit(null, $response->getCacheBody());
-                }
-                if ($response->getResult() == ECacheResult::Miss) {
-                    $this->responsesList[] = new CacheDictionaryGetFieldResponseMiss();
+                    $this->responses[] = new CacheDictionaryGetFieldResponseHit(null, $response->getCacheBody());
+                    $this->valuesDictionary[$fields[$counter]] = $response->getCacheBody();
+                    $counter++;
```
pgautier404, line 997: I'd like to ask you about this counter and about the numRequested param.

src/Cache/_ScsDataClient.php
```diff
@@ -565,18 +565,18 @@ public function dictionaryGetFields(string $cacheName, string $dictionaryName, a
...
             return new CacheDictionaryGetFieldsResponseError($e);
         } catch (Exception $e) {
             return new CacheDictionaryGetFieldsResponseError(new UnknownError($e->getMessage()));
         }
-        if ($dictionaryGetBatchResponse->hasFound()) {
-            return new CacheDictionaryGetFieldsResponseSuccess($dictionaryGetBatchResponse);
+        if ($dictionaryGetFieldsResponse->hasFound()) {
+            return new CacheDictionaryGetFieldsResponseSuccess($dictionaryGetFieldsResponse, fields: $fields);
         }
         return new CacheDictionaryGetFieldsResponseSuccess(null, count($fields));
```
pgautier404, line 581: This should be returning `CacheDictionaryGetResponseHit` and `CacheDictionaryGetResponseMiss` instead of trying to use the `Success` class to represent two different states.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -982,47 +982,46 @@ public function asError(): CacheDictionaryGetFieldsResponseError|null
...
                 } else {
-                    $this->responsesList[] = new CacheDictionaryGetFieldResponseError(new UnknownError(strval($response->getResult())));
+                    $this->responses[] = new CacheDictionaryGetFieldResponseError(new UnknownError(strval($response->getResult())));
                 }
             }
         }
         if (is_null($responses) && !is_null($numRequested)) {
             foreach (range(0, $numRequested - 1) as $ignored) {
-                $this->responsesList[] = new CacheDictionaryGetFieldResponseMiss();
+                $this->responses[] = new CacheDictionaryGetFieldResponseMiss();
```
pgautier404, line 1007: I think all of this is the result of trying to use a `Success` class where we need to be using `Hit` and `Miss` classes instead. I left [a comment](https://github.com/momentohq/client-sdk-php/pull/65/files#r1024591717) on `_ScsDataClient` that lines up with this one.

poppoerika: When I moved `CacheDicitonaryGetBatch` from our .NET incubating repo, I referenced [here](https://github.com/momentohq/client-sdk-dotnet-incubating/blob/main/src/Momento.Sdk.Incubating/Responses/CacheDictionaryGetBatchResponse.cs), which has `Success` and `Error` as `CacheDictionaryGetBatchResponse`.
Sorry if I missed, but are we going to update this response type to `Hit`, `Miss`, and `Error` instead?

pgautier404: You might double check with @cprice404 but I think this should follow the Hit/Miss/Error pattern since we're dealing with a single dictionary value and we can tell if our request was a miss in the corresponding data client code. It should also allow us to get rid of the `$numRequested` parameter and logic in the `CacheDictionaryGetFieldsResponseSuccess` constructor.

cprice404: I think I agree with @pgautier404 , I think this was an oversight when we were switching the .NET SDK over.

Here:

https://github.com/momentohq/client-sdk-dotnet-incubating/blame/main/src/Momento.Sdk.Incubating/Internal/ScsDataClient.cs#L312

That code does not check for the "Missing" case that we got from the server.  My gut is that it should be checking for that case, and should use Hit/Miss/Error pattern instead of success.  That's what we do for DictionaryFetch, ListFech, etc.  So we should fix that both in .NET and PHP.  I created this ticket to capture it for .NET https://github.com/momentohq/client-sdk-dotnet-incubating/issues/47 but yes let's go ahead and make this change here.  @malandis if you know of any reason why DictionaryGetBatch/DictionaryGetFields shouldn't use Hit/Miss/Error pattern let us know.

I didn't review this closely enough to understand why that change would eliminate the need for the numRequested/counter thing but I also agree that it seems like we shouldn't need those, so if they still seem necessary after the switch to Hit/Miss/Error feel free to ping for 👀 .

Sorry for not catching this during the .NET work.

poppoerika: @pgautier404 @cprice404 @malandis 
My understanding is we want to have 4 different response patters:

1. We found a dictionary and all items in the dictionary also found.
2. We found a dictionary but some of the items in the dictionary are missing.
3. We did not find a dictionary.
4. We had an error finding a dictionary.

Cases 1, 3, 4 are straightforward.
1 -> Hit
3 -> Miss
4 -> Error

But how about case 2? Would this be Hit although some of the items are Miss?
So Hit means that finding a dictionary not necessarily the items inside?

cprice404: yes my opinion is that if the Dictionary itself is found, then case 2 is a HIT.

users may then check for individual fields by iterating over the `responses` and seeing which are HIT vs MISS, or they can just use `valueDictionary` and check to see if each key exists.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -980,53 +989,49 @@ public function asError(): CacheDictionaryGetFieldsResponseError|null
...
-                } else {
-                    $this->responsesList[] = new CacheDictionaryGetFieldResponseError(new UnknownError(strval($response->getResult())));
-                }
-            }
-        }
-        if (is_null($responses) && !is_null($numRequested)) {
-            foreach (range(0, $numRequested - 1) as $ignored) {
-                $this->responsesList[] = new CacheDictionaryGetFieldResponseMiss();
+        $counter = 0;
+        parent::__construct();
```
pgautier404, line 1000: Kind of a nit, but the call to the parent constructor should pretty much always be the first line in a function:
```suggestion
        parent::__construct();
        $counter = 0;
```

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -980,53 +989,49 @@ public function asError(): CacheDictionaryGetFieldsResponseError|null
...
+        foreach ($responses->getFound()->getItems() as $response) {
+            if ($response->getResult() == ECacheResult::Hit) {
+                $this->responses[] = new CacheDictionaryGetFieldResponseHit(null, $response->getCacheBody());
+                $this->valuesDictionary[$fields[$counter]] = $response->getCacheBody();
+            } else if ($response->getResult() == ECacheResult::Miss) {
```
pgautier404, line 1005: My IDE is telling me that PHP standards prefer `elseif`.
```suggestion
            } elseif ($response->getResult() == ECacheResult::Miss) {
```

pgautier404:
I left a couple of notes on really minor changes but am good for this to merge once you have a chance to look at those!

pgautier404 (approved):
LGTM!

cprice404 (approved):
🚢

PR #66, opened 2022-11-16
chore: add `setRemoveElement` and tests

poppoerika:
Added `setRemoveElement` since this operation is needed to implement `flush` for `MomentoTaggedCache` (https://github.com/momentohq/laravel-cache/issues/9).
Also, added tests for this operation.


.vscode/launch.json
```diff
@@ -0,0 +1,48 @@
...
+            "cwd": "${workspaceRoot}",
+            "port": 9003,
+            "serverReadyAction": {
+                "pattern": "Development Server \\(http://localhost:([0-9]+)\\) started",
+                "uriFormat": "http://localhost:%s",
+                "action": "openExternally"
+            }
+        }
+    ]
+}
```
cprice404, line 48: i'm not sure if this file should be checked in?

cprice404 (approved):

cprice404:
lgtm other than I'm not sure if we want that vscode file committed to vcs

.vscode/launch.json
poppoerika: 😮‍💨

poppoerika: Will remove that.

cprice404 (approved):

PR #67, opened 2022-11-16
chore: add proxy support

pgautier404:
This contains initial R&D on adding proxy support in SimpleCacheClient. It adds a proxy-aware credential provider to route gRPC calls through a proxy (currently haproxy on layer 4) to the appropriate Momento endpoints. The proxy endpoints are passed in the constructor to the credential provider and the Momento endpoints are parsed out of the JWT token as before.

An haproxy instance configured to send requests on to the Momento backend is required to successfully run the `proxy-example.php` example. You can add the below configuration to your `haproxy.cfg` (after replacing the placeholders with the actual hostnames from your JWT in the `server` directives) to work with the configuration expected by that script.

```
frontend control-plane-fe
  bind momento-control:4443
  option tcplog
  mode tcp
  default_backend control-plane-be

backend control-plane-be
  mode tcp
  server server1 [control-plane-hostname-from-JWT]:443

frontend data-plane-fe
  bind momento-cache:4444
  option tcplog
  mode tcp
  default_backend data-plane-fe

backend data-plane-fe
  mode tcp
  server server1 [data-plane-hostname-from-JWT]:443
```


examples/proxy-example.php
```diff
@@ -0,0 +1,110 @@
...
+$ITEM_DEFAULT_TTL_SECONDS = 60;
+$KEY = "MyKey";
+$VALUE = "MyValue";
+
+// Setup
+$authProvider = new EnvMomentoTokenProxyProvider(
+    "MOMENTO_AUTH_TOKEN",
+    "momento-control:4443",
+    "momento-cache:4444"
+);
```
cprice404, line 29: There is so much other code in this example that this bit gets lost in the noise.  tbh I think if we want to have a proxy example it should probably just have the auth provider instantiation, with more comments about what it's doing and why, and then like one call to "listCaches" or something.  and we should remove the rest of the stuff that's duplicated from the other examples so that their attention is drawn to the important thing.

src/Auth/EnvMomentoTokenProvider.php
```diff
@@ -40,4 +39,14 @@ public function getControlEndpoint(): string
...
+
+    public function getControlProxyEndpoint(): string|null
+    {
+        return null;
+    }
+
+    public function getCacheProxyEndpoint(): string|null
+    {
+        return null;
+    }
```
cprice404, line 51: Left you some input in slack about some other ideas for this interface.

cprice404 (commented):
this is looking great, the way you got it working seems totally reasonable to me.  Just need to finalize our API design.

As an artifact of this ticket I'd appreciate it if you could file a PR against standards and practices repo that includes the haproxy config that you included in your commit message, and a few (language-agnostic) words about how we support reverse proxying and when it might be useful.  Then we can add a bit more detail to the README in this repo and include a link to the s&p doc

README-proxying.md
```diff
@@ -0,0 +1,67 @@
...
+  mode tcp
+  server server1 control.some-control-cell-name.momentohq.com:443
+
+frontend cache-plane-fe
+  bind localhost:4444
+  option tcplog
+  mode tcp
+  default_backend cache-plane-fe
+
+backend cache-plane-fe
```
cprice404, line 38: should this be `-be`

README-proxying.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Configuring the Momento Client
+
+Configuring the Momento client to use the proxy requires the same information, which is passed to the credential
+provider. Using the `EnvMomentoTokenProvider`, which reads the token from an environment variable:
+
+```php
+$authProvider = new EnvMomentoTokenProvider(
+    // name of the environment variable that contains our auth token
+    "MOMENTO_AUTH_TOKEN",
```
cprice404, line 51: does php support named args?

src/Auth/EnvMomentoTokenProvider.php
```diff
@@ -13,17 +12,41 @@ class EnvMomentoTokenProvider implements ICredentialProvider
...
-        $this->cacheEndpoint = $payload->c;
+
+        $endpointArgs = [
+            $controlEndpoint, $cacheEndpoint, $trustedCacheEndpointCertificateName, $trustedControlEndpointCertificateName
+        ];
+        if ($this->anyAreDefined($endpointArgs)) {
+            if (!$this->allAreDefined($endpointArgs)) {
+                throw new InvalidArgumentError(
+                    "If any of controlEndpoint, cacheEndpoint, trustedCacheEndpointCertificateName, or " .
+                    "trustedControlEndpointCertificateName are provided, they must all be.");
```
cprice404, line 39: you can address this in a follow-up pr but why would we disallow the first two being passed by themselves?

cprice404 (approved):
nits, could be handled in separate pr

pgautier404:
I'll address all feedback in a followup PR coming up in just a minute.

README-proxying.md
pgautier404: Good catch, thanks.

README-proxying.md
pgautier404: It does, I'll add them in, thanks.

src/Auth/EnvMomentoTokenProvider.php
pgautier404: Good point.

PR #75, opened 2022-11-19
feat: make Dockerfile more general-purpose

cprice404:
This commit changes the Dockerfile so that we can use a bind mount
to mount the local dir, rather than having to copy all the files
into the image at build time.

This should allow us to re-use this docker image for other purposes;
in the near future we can push it to an ECR repo so that devs
don't have to build it locally at all.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ Check out our SDK [requirements](https://github.com/momentohq/client-sdk-php#req
 You will likely need to run this only once:
 
 ```bash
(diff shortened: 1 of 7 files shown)


poppoerika (dismissed):
Looks good to me! Thank you for doing this, Chris.
It would be so nice to use this as a pre-built image soon :) 
Running `RUN pecl install grpc` is a nightmare.

schwern (commented):
🤷

pgautier404 (approved):
LGTM!

PR #77, opened 2022-11-28
chore: make `$refreshTtl` optional and re-order arguments for list push and dictionary increment 

### LINKED ISSUE #49: change refreshTtl to be optional, default to `true`

### LINKED ISSUE #50: re-order argument list for list push operations; `truncateBackToSize` should appear before ttl options

poppoerika:
- Made `$refreshTtl` optional and default to `true`.
- re-ordered arguments for list push and dictionary increment operations so that `$trunncateFrontToSie`, `$truncateBackToSize`, and `amount` are before `$refreshTtl`.
- updated the tests to reflect the above changes.

This PR needs to get merged before working on #54.

Closes #49 
Closes #50

### DIFF
--- a/src/Cache/SimpleCacheClient.php
+++ b/src/Cache/SimpleCacheClient.php
@@ -32,6 +32,8 @@ use Momento\Cache\CacheOperationTypes\DeleteCacheResponse;
 use Momento\Cache\CacheOperationTypes\ListCachesResponse;
 use Momento\Config\IConfiguration;
 use Momento\Logging\ILoggerFactory;
+use Momento\Requests\CollectionTtl;
+use Momento\Requests\CollectionTtlFactory;
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerInterface;
 
@@ -105,17 +107,17 @@ class SimpleCacheClient implements LoggerAwareInterface
     }
 
     public function listPushFront(
-        string $cacheName, string $listName, string $value, ?int $truncateBackToSize = null, ?bool $refreshTtl = true, ?int $ttlSeconds = null
+        string $cacheName, string $listName, string $value, ?int $truncateBackToSize = null, CollectionTtl $ttl = null
     ): CacheListPushFrontResponse
     {
-        return $this->dataClient->listPushFront($cacheName, $listName, $value, $truncateBackToSize, $refreshTtl, $ttlSeconds);
+        return $this->dataClient->listPushFront($cacheName, $listName, $value, $truncateBackToSize, $ttl);
     }
 
     public function listPushBack(
-        string $cacheName, string $listName, string $value, ?int $truncateFrontToSize = null, ?bool $refreshTtl = true, ?int $ttlSeconds = null
+        string $cacheName, string $listName, string $value, ?int $truncateFrontToSize = null, CollectionTtl $ttl = null
     ): CacheListPushBackResponse
     {
-        return $this->dataClient->listPushBack($cacheName, $listName, $value, $truncateFrontToSize, $refreshTtl, $ttlSeconds);
+        return $this->dataClient->listPushBack($cacheName, $listName, $value, $truncateFrontToSize, $ttl);
     }
 
     public function listPopFront(string $cacheName, string $listName): CacheListPopFrontResponse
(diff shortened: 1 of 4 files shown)


cprice404 (dismissed):
@poppoerika if you want to merge this as an incremental PR it looks good to me, but before we close #49 I wanted to make it conform to the pattern we established in the .NET repo, like this:

https://github.com/momentohq/client-sdk-dotnet-incubating/pull/45

poppoerika:
@cprice404 
As I implemented `CollectionTtl`, I wasn't entirely sure when to use `withRefreshTtlOnUpdates`.
If users want to refresh TTL on updates, it can be achieved by `CollectionTtl::of(10)` since `$refreshTtl = true` by default.
Maybe I'm missing something here.

cprice404:
> @cprice404 As I implemented `CollectionTtl`, I wasn't entirely sure when to use `withRefreshTtlOnUpdates`. If users want to refresh TTL on updates, it can be achieved by `CollectionTtl::of(10)` since `$refreshTtl = true` by default. Maybe I'm missing something here.

Good question.  It's just there for API symmetry, in case people want to write very explicit code.

src/Cache/_ScsDataClient.php
```diff
@@ -175,6 +177,14 @@ private function ttlToMillis(?int $ttl = null): int
         return $ttl * 1000;
     }
 
+    private function returnCollectionTtl($ttl): CollectionTtl
```
cprice404, line 180: can we annotate tye type of the argument here?  I believe it would be `?CollectionTtl`?

src/Requests/CollectionTtl.php
```diff
@@ -0,0 +1,46 @@
+<?php
+declare(strict_types=1);
+
+namespace Momento\Requests;
+
+class CollectionTtl
+{
+    private ?int $ttl;
```
cprice404, line 8: If we are using integers to represent ttls, the variable names need to include the unit for clarity.  I think we were using `ttlSeconds` in the previous code.

cprice404 (commented):
i just have a couple of minor nits and then I'm +1.  We should probably also get Pete to take a quick peek when he's back tomorrow.

cprice404 (dismissed):
✨ 🚢 from me!

but let's let pete take a quick peek

src/Cache/SimpleCacheClient.php
```diff
@@ -143,9 +144,9 @@ public function listErase(string $cacheName, string $listName, ?int $beginIndex
         return $this->dataClient->listErase($cacheName, $listName, $beginIndex, $count);
     }
 
-    public function dictionarySetField(string $cacheName, string $dictionaryName, string $field, string $value, bool $refreshTtl, ?int $ttlSeconds = null): CacheDictionarySetFieldResponse
+    public function dictionarySetField(string $cacheName, string $dictionaryName, string $field, string $value, CollectionTtl $ttl = null): CacheDictionarySetFieldResponse
```
pgautier404, line 147: I'd prefer the `CollectionTtl` parameter be explicitly nullable in all of these functions:

```suggestion
    public function dictionarySetField(string $cacheName, string $dictionaryName, string $field, string $value, ?CollectionTtl $ttl = null): CacheDictionarySetFieldResponse
```

It looks like the null assignment does that implicitly, but I think making it explicit and consistent with the rest of our nullable parameters will make it more readable and maintainable.

src/Cache/_ScsDataClient.php
```diff
@@ -271,18 +281,19 @@ public function listFetch(string $cacheName, string $listName): CacheListFetchRe
     }
 
     public function listPushFront(
-        string $cacheName, string $listName, string $value, bool $refreshTtl, ?int $truncateBackToSize = null, ?int $ttlSeconds = null
+        string $cacheName, string $listName, string $value, ?int $truncateBackToSize = null, CollectionTtl $ttl = null
```
pgautier404, line 284: Same as above: please make all the CollectionTtl parameters in this file explicitly nullable `?CollectionTtl` instead.

src/Requests/CollectionTtl.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->ttlSeconds = $ttlSeconds;
+        $this->refreshTtl = $refreshTtl;
+    }
+
+    public static function fromCacheTtl(): CollectionTtl
+    {
+        return new CollectionTtl(null, true);
+    }
+
+    public static function of(int $ttl): CollectionTtl
```
pgautier404, line 22: I think I'd prefer this parameter to match the constructor and be called `$ttlSeconds` instead, so the unit is explicit in the parameter name (and in the named parameter label). I know I personally get confused about when we're using seconds vs. millis.

pgautier404 (commented):
This looks great! I made a couple small suggestions.

pgautier404 (approved):
This is great, thanks you!

PR #84, opened 2022-12-05
chore: remove collection-specific delete operations

pgautier404:


cprice404 (dismissed):
this looks right to me.  @malandis ?

malandis:
> this looks right to me. @malandis ?

I agree. Only one suggestion: I intend to leave all the test as-is, except replace the specific delete function invocations (`dictionaryDelete`, `setDelete`) with normal `delete`. That way we at least exercise the whole lifecycle (create, read, update, delete) for every data structure.

malandis (approved):

PR #91, opened 2023-01-05
feat: Support SetIfNotExists

rlinehan:
`SetIfNotExists` stores a value for a key if the key is not already present (and returns `Stored`), otherwise it does not store the value (and returns `NotStored`).

Even though it's not really to do with "data structures," this API is only available in caches that also support data structures.

The server side changes for this have not all landed yet, so the tests won't pass in CI. I wanted to get this up for review anyway, but we should hold off merging it until this actually exists for real :grin:.

(Also, I've never written any PHP before this, so a lot of no idea what I'm doing going on here).


tests/Cache/CacheClientTest.php
```diff
@@ -447,6 +447,88 @@ public function testGetTimeout()
...
+    // SetIfNotExists tests
+
+    public function testSetIfNotExists()
+    {
+        $key = uniqid();
+        $value = uniqid();
+
+        $response = $this->client->setIfNotExists($this->TEST_CACHE_NAME, $key, $value);
+        $this->assertNull($response->asError());
+        $response = $response->asStored();
```
pgautier404, line 459: Do you want to an assertion about `$response` not being null here? Also, if you expect this to *not* return `NotStored` you might assert that `asNotStored` returns null.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -357,6 +368,45 @@ class CacheGetResponseError extends CacheGetResponse
...
+    {
+        if ($this->isError()) {
+            return $this;
+        }
+        return null;
+    }
+}
+
+class CacheSetIfNotExistsResponseStored extends CacheSetIfNotExistsResponse {
+}
```
pgautier404, line 399: Our vanilla `set` operatio returns the key and value in the response object. Is there any value in having `CacheSetIfNotExistsResponseStored` behave the same way?

rlinehan: Sure, I can do that for consistency.

tests/Cache/CacheClientTest.php
rlinehan: Whoops, thanks for catching the missing assertion!

src/Cache/CacheOperationTypes/CacheOperationTypes.php
rlinehan: Done in [e454276](https://github.com/momentohq/client-sdk-php/pull/91/commits/e45427622a5cad451d229776518230f4e62e5e6f)

tests/Cache/CacheClientTest.php
rlinehan: Done in [374d3ff](https://github.com/momentohq/client-sdk-php/pull/91/commits/374d3ff8477a745d539875da8e48bad5c2846c88)

pgautier404 (approved):
LGTM!

cprice404 (approved):
seems like there is a CI issue but otherwise 🚢

rlinehan:
Server side changes merged into alpha and CI is passing 🎉

PR #99, opened 2023-03-17
feat: add missing v1 ops

pgautier404:
This commit adds the `types` classes generated from the latest protos and an implementation of the `keysExist` and `keyExists` operations. It also removes a number of unused and/or invalid imports. The rest of the CDT implementations will be coming in as separate PRs to keep them a manageable size for review.


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -459,6 +459,84 @@ class CacheDeleteResponseError extends CacheDeleteResponse
...
+    private array $values = [];
+
+    public function __construct(_KeysExistResponse $response) {
+        parent::__construct();
+        foreach ($response->getExists()->getIterator() as $value) {
+            $this->values[] = (bool)$value;
+        }
+    }
+
+    public function exists() : array {
```
cprice404, line 492: nit, up to you: i am imagining that for most of the SDKs we will want to provide a dictionary accessor for this.  If we do it as an array like this, it's useful if the caller wants to do a `.every` or `.all` check, but otherwise they have to loop over this and their original fields list in parallel to check values and it's kinda janky. 

I feel like as an end user it's a much nicer experience if I can get back a dictionary with the fields as keys.

we can always add this later, so definitely not a blocker.

types/Cache_client/ScsClient.php
```diff
@@ -72,6 +72,48 @@ public function Delete(\Cache_client\_DeleteRequest $argument,
         $metadata, $options);
     }
 
+    /**
+     * @param \Cache_client\_KeysExistRequest $argument input argument
+     * @param array $metadata metadata
+     * @param array $options call options
+     * @return \Grpc\UnaryCall
+     */
+    public function KeysExist(\Cache_client\_KeysExistRequest $argument,
```
cprice404, line 81: everything in this dir is just generated code, yeah?

cprice404 (commented):
just one question, otherwise lgtm

src/Cache/_ScsDataClient.php
```diff
@@ -288,6 +282,33 @@ public function delete(string $cacheName, string $key): CacheDeleteResponse
         return new CacheDeleteResponseSuccess();
     }
 
+    public function keysExist(string $cacheName, array $keys) : CacheKeysExistResponse
+    {
+        try {
+            validateCacheName($cacheName);
+            $keysExistRequest = new _KeysExistRequest();
+            $keysExistRequest->setCacheKeys($keys);
```
poppoerika, line 290: Do we want to validate `$keys` before passing them like [this](https://github.com/momentohq/client-sdk-php/blob/ad4305cd87b8ea9bd74f962239e2f406deac0cd5/src/Utilities/_DataValidation.php#L97)?

types/Cache_client/ScsClient.php
pgautier404: Yes, everything under `types/` is generated.

src/Cache/_ScsDataClient.php
pgautier404: Great callout, I'll add that, thanks!

src/Cache/CacheOperationTypes/CacheOperationTypes.php
pgautier404: I'm fine with that, but I was going off [the SDK API spec](https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083#fbd08eaf1b614a579728f38263dd13df) and it explicitly calls for a list of booleans. For that reason, the golang SDK is returning []bool for this too. I totally agree that a dictionary is a better UX choice. I'm happy to change this up to return a dictionary (which because this is PHP will still be called and typed "an array") but we should think about adding `existsMap` or similar to the gloang SDK as well. And we should obviously correct the spec.

cprice404: we need the list of booleans.  but i think adding a map too is nice.  you can see how we did it in the .NET SDK if you like.  And yes if you wouldn't mind updating the spec that would be great, thank you!!

cprice404: and adding it to go at some point would be great too but it's an additive change so we can do it whenever IMO

pgautier404: Added a "dictionary" accessor in 1a803f98 and updated the spec to reference `existsDictionaryString` and `existsDictionaryBytes` accessors. @cprice404 I opted to omit a "Bool" suffix on those accessors but feel free to change that if you prefer the more verbose version. Because PHP doesn't differentiate, I just added the accessor in this SDK as `existsDictionary`.

poppoerika (approved):
Looks good to me!

PR #104, opened 2023-03-17
chore: update CredentialProvider

### LINKED ISSUE #100: Update CredentialProvider to include string provider and helper methods
The PHP SDK currently only implements the `EnvMomentoTokenProvider` Credential Provider. Other SDKs have introduced a `StringMomentoTokenProvider` that we should port over to PHP. The `fromString` and `fromEnvironmentVariable` static methods that have been introduced in the other SDKs also need to be ported over.

poppoerika:
- Added `StringMomentoTokenProvider`
- Added static functions, `fromEnvironmentVariable` and `fromString` 
- Added some test to test out the above

Closes #100


malandis:
Looks good. Can you document the `StringMomentoProvider` class and `fromEnvironmentVariable` method as well?

pgautier404:
This looks good but there is a lot of duplicated code between the two providers. In other SDKs we had the `EnvMomentoTokenProvider` inherit from `StringMomentoTokenProvider` since the only extra thing it does is read a stering from an environment variable. I think doing the same here would simplify and make this more maintainable. The node SDK is a good model for this pattern.

poppoerika:
> This looks good but there is a lot of duplicated code between the two providers. In other SDKs we had the `EnvMomentoTokenProvider` inherit from `StringMomentoTokenProvider` since the only extra thing it does is read a stering from an environment variable. I think doing the same here would simplify and make this more maintainable. The node SDK is a good model for this pattern.

@pgautier404 
I wanted to discuss this with you, so sounds good!

src/Auth/ICredentialProvider.php
```diff
@@ -1,17 +0,0 @@
-<?php
-declare(strict_types=1);
-
-namespace Momento\Auth;
-
-interface ICredentialProvider
```
cprice404, line 6: why delete the interface?

cprice404 (commented):
i feel like it's probably still a good idea to keep the interface around and use it in the return types from the factory functions, but will defer to Pete.

src/Auth/ICredentialProvider.php
pgautier404: Yeah, we kind of went back and forth on this but I think @cprice404 is right. Keeping the interface around gives us extra flexibility, and that's definitely what we should be returning from the factory methods, @poppoerika.

src/Auth/CredentialProvider.php
```diff
@@ -0,0 +1,30 @@
+<?php
+declare(strict_types=1);
+
+namespace Momento\Auth;
+
+/**
+ * Provides information that the CacheClient needs in order to establish a connection to and authenticate with
+ * the Momento service.
+ */
+abstract class CredentialProvider
```
pgautier404, line 10:
```suggestion
abstract class CredentialProvider implements ICredentialProvider
```

We'll of course need to add the `ICredentialProvider` file and class back in.

src/Auth/EnvMomentoTokenProvider.php
```diff
@@ -27,42 +23,17 @@ public function __construct(
...
     }
 
-    public function getTrustedCacheEndpointCertificateName(): string|null
+    /**
+     * Convenience method for reading and parsing a JWT token stored as an environment variable.
+     * @param string $envVariableName Name of the environment variable that contains the JWT token.
+     * @return EnvMomentoTokenProvider
+     * @throws InvalidArgumentError
+     */
+    public static function fromEnvironmentVariable(string $envVariableName): EnvMomentoTokenProvider
```
pgautier404, line 35: This function should be in the `CredentialProvider` class and should return `ICredentialProvider` type.

src/Auth/StringMomentoTokenProvider.php
```diff
@@ -0,0 +1,74 @@
...
+        return $this->trustedCacheEndpointCertificateName;
+    }
+
+    /**
+     * Convenience method for reading and parsing a JWT token stored as a string.
+     * @param string $authToken The JWT token.
+     * @return StringMomentoTokenProvider
+     * @throws InvalidArgumentError
+     */
+    public static function fromString(string $authToken): StringMomentoTokenProvider
```
pgautier404, line 70: Please move this function to `CredentialProvider` and mark it as returning an `ICredentialProvider`.

src/Cache/CacheClient.php
```diff
@@ -48,11 +48,11 @@ class CacheClient implements LoggerAwareInterface
 
     /**
      * @param IConfiguration $configuration
-     * @param ICredentialProvider $authProvider
+     * @param CredentialProvider $authProvider
      * @param int $defaultTtlSeconds
      */
     public function __construct(
-        IConfiguration $configuration, ICredentialProvider $authProvider, int $defaultTtlSeconds
+        IConfiguration $configuration, CredentialProvider $authProvider, int $defaultTtlSeconds
```
pgautier404, line 55:
```suggestion
        IConfiguration $configuration, ICredentialProvider $authProvider, int $defaultTtlSeconds
```

pgautier404: Sorry about the runaround, but we'll want these types to be `ICredentialProvider` pretty much everywhere in the clients and grpc managers.

poppoerika:
@pgautier404 @cprice404 
Thank you for the feedback! 
After talking to Pete, it makes sense to keep `ICredentialProvider` and implement it in `CredentialProvider` to add flexibility to our interface.
I added back `ICredentialProvider` and updated `CredentialProvier` accordingly. 
Also updated the examples so that we use `CacheClient` and `CredentialProvider::fromEnvironmentVariable`.
Ah, also updated `CredentialProvider` return type to `ICredentialProvier` type.

src/Cache/Psr16CacheClient.php
```diff
@@ -26,15 +26,15 @@ class Psr16CacheClient implements CacheInterface
 
     /**
      * @param IConfiguration $configuration
-     * @param ICredentialProvider $authProvider
+     * @param CredentialProvider $authProvider
```
pgautier404, line 29:
```suggestion
     * @param ICredentialProvider $authProvider
```

src/Cache/Psr16CacheClient.php
```diff
@@ -26,15 +26,15 @@ class Psr16CacheClient implements CacheInterface
...
      * @param int|null $defaultTtlSeconds
      * @param bool|null $throwExceptions
      */
     public function __construct(
-        IConfiguration      $configuration,
-        ICredentialProvider $authProvider,
-        ?int                $defaultTtlSeconds,
-        ?bool               $throwExceptions = null
+        IConfiguration     $configuration,
+        CredentialProvider $authProvider,
```
pgautier404, line 35:
```suggestion
        ICredentialProvider $authProvider,
```

src/Auth/CredentialProvider.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public abstract function getTrustedCacheEndpointCertificateName(): string|null;
+
+    /**
+     * Convenience method for reading and parsing a JWT token stored as a string.
+     * @param string $authToken The JWT token.
+     * @return StringMomentoTokenProvider
+     * @throws InvalidArgumentError
+     */
+    public static function fromString(string $authToken): StringMomentoTokenProvider
```
pgautier404, line 26:
```suggestion
    public static function fromString(string $authToken): ICredentialProvider
```

src/Auth/CredentialProvider.php
```diff
@@ -0,0 +1,41 @@
...
+        return new StringMomentoTokenProvider($authToken);
+    }
+
+    /**
+     * Convenience method for reading and parsing a JWT token stored as an environment variable.
+     * @param string $envVariableName Name of the environment variable that contains the JWT token.
+     * @return EnvMomentoTokenProvider
+     * @throws InvalidArgumentError
+     */
+    public static function fromEnvironmentVariable(string $envVariableName): EnvMomentoTokenProvider
```
pgautier404, line 37:
```suggestion
    public static function fromEnvironmentVariable(string $envVariableName): ICredentialProvider
```

pgautier404 (approved):
LGTM!

PR #105, opened 2023-03-17
chore: rename SimpleCacheClient -> CacheClient

pgautier404:


poppoerika (approved):
Looks good to me 🍀

src/Cache/CacheClient.php
```diff
@@ -38,7 +37,7 @@
 use Psr\Log\LoggerAwareInterface;
```
cprice404, line 37: any idea what all the qodana stuff below is?  i feel like we should either disable that or add whatever it is doing to our CI linting?  doesn't have to be in this PR but it does make it kinda hard to review :)

cprice404 (approved):
would love to sort the qodana thing but not a blocker for this PR

src/Cache/CacheClient.php
pgautier404: No idea, but these are clearly false positives. I downloaded the binary and it's definitely going to take some digging to see what's up with this thing. I'll go ahead and disable it for now until I have some time to take a closer look.

PR #116, opened 2023-03-24
feat: logging fixup

pgautier404:


src/Logging/ILoggerFactory.php
```diff
@@ -4,8 +4,11 @@
...
 
 use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
 
 interface ILoggerFactory
 {
-    public function getLogger(?string $name): LoggerInterface;
+    public function getLogger(?string $name=null): LoggerInterface;
+
+    public function getLogLevel(): string|null;
```
cprice404, line 13: i'm not sure you want to put this here.

i am not familiar with the lay of the land in PHP logging but in other languages, logging frameworks will have their own way to configure the level, out-of-band from our code.  some may not even have levels.  So you really want the level to be an implementation detail for the individual logger, not something that we force through the logger factory.  e.g. see: 

https://github.com/momentohq/client-sdk-go/blob/main/examples/logging-example/momento_logrus/logrus.go

src/Logging/StderrLoggerFactory.php
```diff
@@ -4,15 +4,19 @@
...
+use Psr\Log\LogLevel;
 
 /**
- * Factory class for returning a StderrEchoLogger instance.
+ * Factory class for returning a StderrLogger instance.
  */
-class StderrLoggerFactory implements ILoggerFactory
+class StderrLoggerFactory extends LoggerFactoryBase
 {
+    public function __construct(?string $logLevel=LogLevel::INFO) {
```
cprice404, line 14: it's fine to require log level here if it's appropriate for this logger implementation.  doesn't mean it needs to be in the factory interface.

take all this with a grain of salt, you're more familiar with the PHP ecosystem than I am, but the advice i'm giving here was definitely the way to go for the other languages I've touched so far.

src/Logging/ILoggerFactory.php
pgautier404: Totally right, that wasn't supposed to be there, thanks. Fixed in a0f0e7e3.

src/Logging/StderrLoggerFactory.php
pgautier404: Yep, it's here and defaulted to INFO as an implementation detail of this class only.

src/Logging/StderrLogger.php
```diff
@@ -4,21 +4,40 @@
...
+    private array $logLevelMap = [
+        LogLevel::EMERGENCY => 80,
+        LogLevel::ALERT => 70,
+        LogLevel::CRITICAL => 60,
+        LogLevel::ERROR => 50,
+        LogLevel::WARNING => 40,
+        LogLevel::NOTICE => 30,
+        LogLevel::INFO => 20,
+        LogLevel::DEBUG => 10,
+    ];
```
malandis, line 27: Wow

malandis (approved):

PR #119, opened 2023-03-27
fix: qodana updates

pgautier404:
This commit introduces some code fixes pointed out by the qodana tool. It also reconfigures qodana to suppress some false positives with as much granularity as the tool allows (which is not a lot).


src/Auth/AuthUtils.php
```diff
@@ -23,7 +23,7 @@ public static function parseAuthToken(string $authToken): object
         }
 
         try {
-            list($header, $payload, $signature) = $exploded;
+            $payload = $exploded[1];
```
poppoerika, line 26: What does this do?

src/Cache/Internal/ScsDataClient.php
```diff
@@ -206,13 +206,13 @@ public function set(string $cacheName, string $key, string $value, int $ttlSecon
...
             );
-            $response = $this->processCall($call);
+            $this->processCall($call);
         } catch (SdkError $e) {
             return new SetError($e);
         } catch (Exception $e) {
             return new SetError(new UnknownError($e->getMessage()));
         }
-        return new SetSuccess($response, $key, $value);
+        return new SetSuccess();
```
poppoerika, line 215: Sorry if I missed in your previous PRs, but we don't take any params to create an instance of the `SetSuccess` anymore?

src/Auth/AuthUtils.php
pgautier404: I only use the `$payload`, so this gets rid of a warning about the other two variables being unused.

src/Cache/Internal/ScsDataClient.php
pgautier404: No, we don't.

poppoerika (approved):

PR #132, opened 2023-05-05
feat: Add a prototype async get method

nand4011:
Add an async get method that separates the grpc call from interpreting the results. This is significantly faster depending on how the method is used.

Add a temporary test that runs the sync and async get methods in groups of 50 ten times and prints how long it takes. The async method is ~24x faster in this test.

If we wanted to reduce effort on adding async methods, we could make the async versions return a callable instead of a special future class, but then we would lose type hinting.

We could also hide the async methods entirely by making the `isHit()`/`asHit()` methods implicitly call the callable instead of providing it to the user. This would maintain compatibility with all of the tests except for the toString() tests. The downside is that it wouldn't be obvious that the methods were asynchronous unless the user reads the documentation.


cprice404:
nice job, thanks!

pgautier404:
Obsoleted by https://github.com/momentohq/client-sdk-php/pull/157.

PR #137, opened 2023-05-25
feat: add idle data client wrapper

pgautier404:
This commit adds support for an idle client timeout to address the issue of dropped connections in lambda environments.


cprice404 (approved):
LGTM!

cprice404:
oh, looks like you may have to fix the auth token support before we can merge this though.

cprice404 (approved):

PR #140, opened 2023-05-26
fix: explicitly close gRPC channel when cycling idle client

cprice404:
Prior to this commit we didn't ever have an explicit call to close
the gRPC channel; in cases where we are cycling the client out
due to reaching the idle timeout, this could leave open sockets /
file descriptors behind.  This commit updates the code to explicitly
close the channel when replacing the client.


cprice404:
@pgautier404 I haven't tested this thoroughly, would appreciate if you could test it before we merge.

pgautier404 (approved):
LGTM!

PR #149, opened 2023-06-03
feat: Implement set union operation

GrahamCampbell:
Adds support for adding multiple items to a set in a single request, implementing using the existing operation already implemented in the underlying gRPC client.


GrahamCampbell:
Tests failing because `TEST_AUTH_TOKEN` is not available to PRs from forks I presume.

cprice404:
> Tests failing because `TEST_AUTH_TOKEN` is not available to PRs from forks I presume.

Yes. We have some plans for how to fix this issue in the future but haven't gotten there yet.

PR #153, opened 2023-06-03
fix: use $_SERVER instead of getenv

cprice404:
This commit changes our token provider code to use $_SERVER
instead of `getenv`. `getenv` is not threadsafe and can cause
security issues.

### DIFF
--- a/src/Auth/EnvMomentoTokenProvider.php
+++ b/src/Auth/EnvMomentoTokenProvider.php
@@ -19,7 +19,7 @@ class EnvMomentoTokenProvider extends StringMomentoTokenProvider
         ?string $trustedCacheEndpointCertificateName = null
     )
     {
-        $authToken = getenv($envVariableName);
+        $authToken = $_SERVER[$envVariableName];
         if ($authToken === false || isNullOrEmpty($authToken)) {
             throw new InvalidArgumentError("Environment variable $envVariableName is empty or null.");
         }
--- a/src/Auth/EnvMomentoTokenProvider.php
+++ b/src/Auth/EnvMomentoTokenProvider.php
@@ -19,7 +19,7 @@ class EnvMomentoTokenProvider extends StringMomentoTokenProvider
         ?string $trustedCacheEndpointCertificateName = null
     )
     {
-        if (! array_key_exists($envVariableName, $_SERVER)) {
+        if ((! array_key_exists($envVariableName, $_SERVER)) || $_SERVER[$envVariableName] == null) {
             throw new InvalidArgumentError("Environment variable $envVariableName is empty or null.");
         }
         $authToken = $_SERVER[$envVariableName];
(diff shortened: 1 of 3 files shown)


src/Auth/EnvMomentoTokenProvider.php
```diff
@@ -19,7 +19,7 @@ public function __construct(
         ?string $trustedCacheEndpointCertificateName = null
     )
     {
-        $authToken = getenv($envVariableName);
+        $authToken = $_SERVER[$envVariableName];
```
GrahamCampbell, line 22: This will crash if the key does not exist in the server super global. One should add `?? false`.

```
$authToken = $_SERVER[$envVariableName] ?? false;
```

GrahamCampbell (changes requested):

tests/Cache/MomentoTokenProviderTest.php
```diff
@@ -24,9 +24,9 @@ class MomentoTokenProviderTest extends TestCase
     public function setUp(): void
     {
         try {
```
GrahamCampbell, line 26: I'd replace lines 26-32 with:

```php
if (!isset($_SERVER[self::AUTH_TOKEN_NAME])) {
    throw new RuntimeException(
        "Integration tests require TEST_AUTH_TOKEN environment variable"
    );
}

$this->authToken = $_SERVER[self::AUTH_TOKEN_NAME];
```

GrahamCampbell (changes requested):

src/Auth/EnvMomentoTokenProvider.php
```diff
@@ -19,10 +19,10 @@ public function __construct(
         ?string $trustedCacheEndpointCertificateName = null
     )
     {
-        $authToken = getenv($envVariableName);
-        if ($authToken === false || isNullOrEmpty($authToken)) {
+        if (! array_key_exists($envVariableName, $_SERVER)) {
```
GrahamCampbell, line 22: array key exists is not correct to use here, since it will pass if the value is present but null?

GrahamCampbell (changes requested):

src/Auth/EnvMomentoTokenProvider.php
```diff
@@ -19,10 +19,10 @@ public function __construct(
         ?string $trustedCacheEndpointCertificateName = null
     )
     {
-        $authToken = getenv($envVariableName);
-        if ($authToken === false || isNullOrEmpty($authToken)) {
+        if ((! array_key_exists($envVariableName, $_SERVER)) || $_SERVER[$envVariableName] == null) {
```
GrahamCampbell, line 22: Can we do?

```php
if (isNullOrEmpty($_SERVER[$envVariableName] ?? null)) {
```

pgautier404 (approved):

PR #158, opened 2023-06-05
feat: add configurability for number of gRPC channels

cprice404:
For users who may end up with over 100 concurrent requests in flight,
we need to be able to support multiple gRPC channels. This commit adds
a configuration setting to allow users to specify the desired number
of channels.

This most likely won't doing anything useful until we add the config
flag to disable persistent TCP connections in the gRPC channel
constructor, because we will just end up re-using the same channel
multiple times.

### DIFF
--- a/src/Cache/CacheClient.php
+++ b/src/Cache/CacheClient.php
@@ -54,7 +54,12 @@ class CacheClient implements LoggerAwareInterface
     protected ILoggerFactory $loggerFactory;
     protected LoggerInterface $logger;
     private ScsControlClient $controlClient;
-    private IdleDataClientWrapper $dataClientWrapper;
+
+    /**
+     * @var IdleDataClientWrapper[]
+     */
+    private array $dataClients;
+    private int $nextDataClientIndex = 0;
 
     /**
      * @param IConfiguration $configuration Configuration to use for transport.
@@ -77,7 +82,10 @@ class CacheClient implements LoggerAwareInterface
                 $defaultTtlSeconds
             );
         };
-        $this->dataClientWrapper = new IdleDataClientWrapper($dataClientFactory, $this->configuration);
+        $this->dataClients = [];
+        for ($i = 0; $i < $configuration->getTransportStrategy()->getGrpcConfig()->getNumGrpcChannels(); $i++) {
+            array_push($this->dataClients, new IdleDataClientWrapper($dataClientFactory, $this->configuration));
+        }
     }
 
     /**
@@ -171,7 +179,7 @@ class CacheClient implements LoggerAwareInterface
      */
     public function setAsync(string $cacheName, string $key, string $value, int $ttlSeconds = 0): ResponseFuture
     {
-        return $this->dataClientWrapper->getClient()->set($cacheName, $key, $value, $ttlSeconds);
+        return $this->getNextDataClient()->set($cacheName, $key, $value, $ttlSeconds);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/Cache/CacheClient.php
```diff
@@ -750,6 +758,13 @@ public function setLength(string $cacheName, string $setName): SetLengthResponse
...
     public function setRemoveElement(string $cacheName, string $setName, string $element): SetRemoveElementResponse
     {
-        return $this->dataClientWrapper->getClient()->setRemoveElement($cacheName, $setName, $element);
+        return $this->getNextDataClient()->setRemoveElement($cacheName, $setName, $element);
+    }
+
+    private function getNextDataClient(): ScsDataClient
+    {
+        $client = $this->dataClients[$this->nextDataClientIndex]->getClient();
+        $this->nextDataClientIndex = ($this->nextDataClientIndex + 1) % sizeof($this->dataClients);
```
GrahamCampbell, line 767: `sizeof` is a deprecated alias of `count`

src/Cache/CacheClient.php
```diff
@@ -163,7 +171,7 @@ public function deleteCache(string $cacheName): DeleteCacheResponse
      */
     public function set(string $cacheName, string $key, string $value, int $ttlSeconds = 0): SetResponse
     {
-        return $this->dataClientWrapper->getClient()->set($cacheName, $key, $value, $ttlSeconds);
+        return $this->getNextDataClient()->set($cacheName, $key, $value, $ttlSeconds);
```
GrahamCampbell, line 174: Seems like a fair approach, to just round robin between the clients. Should be reasonably fair, unless there is some kind of blockage on a connection, though for my use case, I'd jam everything in at the start, so there isn't much opportunity to weight choice towards connections with fewer streams/lower latency.

GrahamCampbell (commented):
> This most likely won't doing anything useful until we add the config flag to disable persistent TCP connections

Fair. Probably there should be validation to mark the two as incompatible with each other. That is, one must disable persistent connections in order to set a non-default number of number of channels.

cprice404:
> > This most likely won't doing anything useful until we add the config flag to disable persistent TCP connections
> 
> Fair. Probably there should be validation to mark the two as incompatible with each other. That is, one must disable persistent connections in order to set a non-default number of number of channels.

yeah we will add some validation to this setting once the other setting lands.

src/Cache/CacheClient.php
cprice404: yeah, round robin is the way we've been doing it in the other languages so far.

src/Cache/CacheClient.php
cprice404: gtk thanks

src/Cache/CacheClient.php
```diff
@@ -77,7 +83,16 @@ public function __construct(
                 $defaultTtlSeconds
             );
         };
-        $this->dataClientWrapper = new IdleDataClientWrapper($dataClientFactory, $this->configuration);
+        $this->dataClients = [];
+
+        $numGrpcChannels = $configuration->getTransportStrategy()->getGrpcConfig()->getNumGrpcChannels();
+        $forceNewChannels = $configuration->getTransportStrategy()->getGrpcConfig()->getForceNewChannel();
+        if (($numGrpcChannels > 0) && (! $forceNewChannels)) {
```
pgautier404, line 90:
```suggestion
        if (($numGrpcChannels > 1) && (! $forceNewChannels)) {
```

pgautier404 (approved):
LGTM!

PR #159, opened 2023-06-05
feat: add config setting to force new channel

pgautier404:
This commit exposes a configuration option that allows users to override the default persistence setting on gRPC channel creation. By default, channels are "persistent"-- given the same set of parameters passed to the constructor, the same underlying grpc channel will be returned. Users can now call `withForceNewChannel(true)` on their Configuration object to change that setting so that a new and separate underlying grpc channel will be created and returned.


cprice404 (commented):
lgtm other than the minor suggestion re: possibly only exposing this at the grpc config level for now. Also curious if you were able to verify this via`lsof` yet or if I should try that out in my branch.

cprice404 (approved):

PR #166, opened 2023-06-06
fix: allow users to override PSR-16 cache name

pgautier404:


src/Cache/Psr16CacheClient.php
```diff
@@ -29,20 +30,23 @@ class Psr16CacheClient implements CacheInterface
...
      * @param bool|null $throwExceptions
+     * @param string $cacheName
      */
     public function __construct(
         IConfiguration      $configuration,
         ICredentialProvider $authProvider,
         ?int                $defaultTtlSeconds,
-        ?bool               $throwExceptions = null
+        ?bool               $throwExceptions = null,
+        string              $cacheName = self::DEFAULT_CACHE_NAME
```
GrahamCampbell, line 40: I would make this nullable and default to null, then do `?? self::DEFAULT_CACHE_NAME` on line 48 so that callers can pass null to indicate they want the default.

GrahamCampbell (changes requested):

GrahamCampbell (approved):

cprice404 (approved):

GrahamCampbell:
A breaking change, but my preference would also be to not make the create cache call in the constructor. Providing a static factory method that does the create call and then returns the instance would be my preference.

cprice404:
> A breaking change, but my preference would also be to not make the create cache call in the constructor. Providing a static factory method that does the create call and then returns the instance would be my preference.

i'm okay with this if you are, @pgautier404 !

pgautier404:
I did some playing around with a static factory but wasn't happy with the shape of what I came up with. I'm going to go ahead and merge this as-is to make something more usable available to users and will revisit when I have some time to take a closer look.

PR #174, opened 2023-11-06
feat: add `close` method for explicitly closing client

cprice404:
This is intended to help us determine whether or not a persistent gRPC connection that gets into a bad state can be explicitly closed and re-opened to restore the environment to a healthy state.


cprice404:
cc: @GrahamCampbell

pgautier404 (approved):

PR #176, opened 2024-01-03
feat: add topic poc

rishtigupta:


composer.json
```diff
@@ -45,7 +46,8 @@
         "google/protobuf": "3.22.3",
         "grpc/grpc": "1.52.0",
         "psr/log": "^1.1 || ^2.0 || ^3.0",
-        "psr/simple-cache": "^1.0.1 || ^3.0"
+        "psr/simple-cache": "^1.0.1 || ^3.0",
+        "ext-json": "*"
```
pgautier404, line 50: What is this for?

rishtigupta: Oh, that was added to print the objects in json format easily

rishtigupta: I used it for debugging. Don't need itnow, will remove it.

pgautier404: Can you go ahead and remove it please? This is a system extension, not a module, and as far as I know it's not necessary. For example, the credential provider has been using `json_encode` and `json_decode` just fine without it. If it did solve a particular problem you were having, let me know and we can discuss.

PR #177, opened 2024-02-07
feat: add getBatch and setBatch

rishtigupta:
## PR Description:
- Add getBatch and setBatch api support

## References:
[Design Doc (Notion)](https://www.notion.so/momentohq/setBatch-getBatch-e7c0c8b34e3c4ca7ac29ea3a79b2d9ac)


types/.DS_Store
pgautier404, line 1: Please remove.

cprice404:
Why are there a bunch of files like `TopicPermission` deleted from this PR?

src/Cache/Internal/ScsDataClient.php
```diff
@@ -30,6 +32,7 @@
 use Cache_client\_SetUnionRequest;
 use Cache_client\ECacheResult;
 use Exception;
+use Grpc\ServerStreamingCall;
```
cprice404, line 35: Where is this used?

rishtigupta:
> Why are there a bunch of files like `TopicPermission` deleted from this PR?

I had to recreate the protos from the [client-types-php](https://github.com/momentohq/client-types-php) to get the latest protos in the sdk.

src/Cache/Internal/ScsDataClient.php
```diff
@@ -1093,6 +1102,87 @@ function () use ($call): SetRemoveElementResponse {
...
+            $getRequests = [];
+            foreach ($keys as $key) {
+                $getRequest = new _GetRequest();
+                $getRequest->setCacheKey($key);
+                $getRequests[] = $getRequest;
+            }
+
+            $getBatchRequest = new _GetBatchRequest();
+            $getBatchRequest->setItems($getRequests);
+            $call = $this->grpcManager->client->GetBatch($getBatchRequest, ['cache' => [$cacheName]], ['timeout' => $this->timeout]);
```
cprice404, line 1123: so this is a streaming call, right?

src/Cache/Internal/ScsDataClient.php
```diff
@@ -1093,6 +1102,87 @@ function () use ($call): SetRemoveElementResponse {
...
+        } catch (SdkError $e) {
+            return ResponseFuture::createResolved(new GetBatchError($e));
+        } catch (Exception $e) {
+            return ResponseFuture::createResolved(new GetBatchError(new UnknownError($e->getMessage(), 0, $e)));
+        }
+
+        return ResponseFuture::createPending(
+            function () use ($call): GetBatchResponse {
+                try {
+                    $responses = $call->responses();
```
cprice404, line 1133: we really don't have to do anything special in order to aggregate / loop over the streaming responses? This function just does all of that?

tests/Cache/CacheClientTest.php
```diff
@@ -2333,4 +2333,77 @@ public function testSetDelete_SetExists_HappyPath()
         $this->assertNull($response->asError());
         $this->assertNotNull($response->asMiss(), "Expected a miss but got $response.");
     }
+
+    public function testGetBatch_HappyPath()
```
pgautier404, line 2337: Could you add some testing for cache misses as well, please? I'd like to test that getBatch is correctly handling and representing them in the `results()` call as well as the `values()` one.

src/Cache/Internal/ScsDataClient.php
rishtigupta: The $calls->responses return the generator function. We iterate over the generator function inside the Success call and create our final result and values

cprice404: Okay, I think I get that. Can you add some tests for the Async variants too then? I'd like to see what it looks like to consume them so I can reason about when the loop happens relative to the `wait`.

pgautier404: It looks like we're missing a call to `$call->getStatus()` to make sure the call was actually successful. I'm able to pass a nonexistent cache name to both `getBatch` and `setBatch` and they appear to succeed.

rishtigupta: We cannot wait after the call has been created since the responses are still streaming. I can create a status function in the success call to see what was the call status after the wait was performed on it?

cprice404: Re: `wait`: okay; maybe I just need to see how a user would consume our future then.

cprice404: and maybe i'm talking about the `wait` on our `Future` class, not one from grpc

src/Cache/Internal/ScsDataClient.php
rishtigupta: yes

src/Cache/Internal/ScsDataClient.php
rishtigupta: I am doing a wait inside the getBtach function in CacheClient: https://github.com/momentohq/client-sdk-php/pull/177/files#diff-faee61621cf99402304d5ff4af286157f07f3fb59e14fefcd672fd0a90a79e84R1219. 
Is this what you're looking for?

cprice404: yeah but I would still like to see what it looks like for an end user who wants to use Async, and I think it's worth having some basic test coverage for those Asyncs

cprice404:
> > Why are there a bunch of files like `TopicPermission` deleted from this PR?
> 
> I had to recreate the protos from the [client-types-php](https://github.com/momentohq/client-types-php) to get the latest protos in the sdk.

That's fine, but I wouldn't have expected those files to get deleted entirely... did they get moved to a different path or something? I would expect stuff like TopicPermission to still exist in the generated code?

src/Cache/Internal/ScsDataClient.php
rishtigupta: removed the import

src/Utilities/_DataValidation.php
```diff
@@ -148,6 +148,14 @@ function validateElements(array $elements): void
     }
 }
 
+if (!function_exists('validateItems')) {
+    function validateItems(array $items): void
+    {
+        $keys = array_keys($items);
+        validateKeys($keys);
```
pgautier404, line 155: Kind of picky, but rather than adding a new function here can you just move this logic to the `setBatch()` call in the data client, replacing `validateItems($items)` with `validateKeys(array_keys($items))`? Because this method doesn't actually do any validation of the items in the array, there's really no compelling reason for the indirection.

tests/Cache/CacheClientTest.php
```diff
@@ -2333,4 +2336,118 @@ public function testSetDelete_SetExists_HappyPath()
...
+        $success = $getBatchResponse->asSuccess();
+        $this->assertNull($getBatchResponse->asError());
+        $this->assertNotNull($success, "Expected a success but got: $getBatchResponse");
+
+        $responses = $success->values();
+        $this->assertCount(sizeof($keys), $responses);
+        $this->assertEquals($responses, $expectedValues);
+    }
+
+    public function testGetBatchSetBatch()
```
pgautier404, line 2414: Could you please add a copy of this test that uses the async methods to provide an example of how those should be used? I have it on my to do list to the same for the scalar async methods.

src/Cache/CacheClient.php
```diff
@@ -1168,4 +1170,104 @@ private function getNextDataClient(): ScsDataClient
...
+     * <code>$response = $responseFuture->wait();
+     * }</code>
+     * The response represents the result of the get operation and stores the
+     * retrieved value. This result is resolved to a type-safe object of one of
+     * the following types:<br>
+     * * GetBatchSuccess<br>
+     * * GetBatchError<br>
+     * Pattern matching can be to operate on the appropriate subtype:<br>
+     * <code>if ($success = $response->asSuccess()) {<br>
+     * &nbsp;&nbsp;$value = $success->getResponses();<br>
```
pgautier404, line 1190: This should be either `$results = $success->results();` or `$values = $success->values();` (or maybe both), but I would like to double check with @cprice404 that that's ok for the final naming. Actually, the docs we have for sources of truth for naming here aren't exactly clear, so we should have a quick discussion about the desired outcome and then feed that back into the docs.

pgautier404 (approved):
LGTM!

PR #178, opened 2024-02-08
chore: remove composer lock

pgautier404:


GrahamCampbell:
Why? It's good OSS supply chain security to commit this. Other momento libs also commit their lock files: https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-web/package-lock.json.

src/Cache/Psr16CacheClient.php
```diff
@@ -111,7 +111,7 @@ public function getLastError(bool $clear_error = true): ?CacheException
     /**
      * @inheritDoc
      */
-    public function get($key, $default = null)
+    public function get(string $key, mixed $default = null): mixed
```
GrahamCampbell, line 114: This change will break things for older psr-16 cache versions, and is not needed for newer versions.

GrahamCampbell (changes requested):

src/Cache/Psr16CacheClient.php
GrahamCampbell: See https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters.

pgautier404: Our PHP 8 tests were failing because the declaration wasn't compatible: https://github.com/momentohq/client-sdk-php/actions/runs/7833550796/job/21374774900. It seems like fixing it for 7 breaks it for 8 and vice versa. The tests are indicating it is needed for PHP 8.0. Any advice on making this mutually compatible would definitely be appreciated :-)

The larger issue that motivated experimenting with removing `composer.lock` was also due to the PSR-16 interface, as we couldn't get composer to calculate a set of dependencies that is valid for both PHP 7 and PHP 8.x.

pgautier404: Ah, sorry, I missed your previous comment. Thanks for the link, I'll check it out!

GrahamCampbell: Probably the issue you have is that supporting PHP 7.4 caused you a pain in the ass because it doesn't have the mixed type.

GrahamCampbell: You could do some jank and conditionally define the class depending on PHP version. Demo:

1. https://3v4l.org/2sCRB
2. https://3v4l.org/n72t8

You can see that PHP 8 works fine with both the interface with and without types, but not PHP 7.

pgautier404: Thanks, Graham, I appreciate the insight.

composer.json
```diff
@@ -44,7 +44,7 @@
         "firebase/php-jwt": "^6.3",
         "google/protobuf": "3.22.3",
         "grpc/grpc": "1.52.0",
-        "psr/log": "^1.1 || ^2.0 || ^3.0",
```
GrahamCampbell, line 47: if we leave this in, i won't be able to user newer versions of the sdk. i need psr log v3.

GrahamCampbell (changes requested):

composer.json
GrahamCampbell: if you want to get a php 7.4 compatible lock file, you can use the platform config to make composer pretend it's a particular version of php. https://getcomposer.org/doc/06-config.md#platform. for example: https://github.com/GrahamCampbell/GitHub-Notifications/blob/4.1/composer.json#L51.

pgautier404: @GrahamCampbell Understood, this isn't intended to be merged. I was trying to see if there was even a theoretical way for me to have a `composer.lock` acceptable to both 7 and 8, and it looks like it is not.

The conditional class definition worked fine and tests are passing with all versions now. We'll be migrating that change into another PR in process, and I'm just doing experiments on this branch.

GrahamCampbell: setting the platform to 7.4.15 would work on php 7.4 through 8.3, but it would not let you run the tests on the newer libs.

pgautier404: @GrahamCampbell is the change in 438eab9 what you were suggesting? If so, it's producing another mole to whack as it's causing the PHP 8 code in the conditional class definition I just added to [fail for PHP 8 again](https://github.com/momentohq/client-sdk-php/actions/runs/7836168717/job/21383177523?pr=178) (as it's pulling in psr/log 1.1.4). If you had something different in mind, could you please elaborate?

GrahamCampbell: The conditional class generation is incorrect. Both cases should never have param types, and only the PHP case should have the return types. PHP users can install any of the 3 versions of the package, so the PHP version doesn't tell you what lib version. If v1 is installed, having param types is illegal.

pgautier404: @GrahamCampbell I see, thanks. I don't have a tremendous understanding of the implications of that platform setting apart from, as you mentioned, it's going to favor older library versions meaning we won't be testing more recent versions in CI. I definitely want to be certain that this can't cause breaking changes for existing users of the SDK. This config change and the new composer.lock don't run the risk of downgrading packages that were installed previously, do they?

GrahamCampbell: no. the platform config and composer.lock file are not inherited or visible at all to package consumers. they are only relevant to people directly working on this repo.

cprice404 (approved):

composer.json
pgautier404: Perfect, thank you.

pgautier404:
Closing experimental branch. Relevant composer platform update and PSR-16 fixes have been committed in #177.

PR #183, opened 2024-05-17
feat: Support fractional TTLs via floats/doubles

GrahamCampbell:
Maintains a BC interface (assuming we say extending the class is not a valid use case), enabling ms-precision of TTLs.

FYI 1, float nearly always means "double" in PHP land - the type depends on the CPU.
FYI 2, due to needing to support PHP 7.4, we can't use union types, only type-doc.

### DIFF
--- a/src/Cache/Internal/ScsDataClient.php
+++ b/src/Cache/Internal/ScsDataClient.php
@@ -189,7 +189,7 @@ class ScsDataClient implements LoggerAwareInterface
     /**
      * @param int|float|null $ttl
      */
-    private function ttlToMillis(?int $ttl = null): int
+    private function ttlToMillis($ttl = null): int
     {
         if (!$ttl) {
             $ttl = $this->defaultTtlSeconds;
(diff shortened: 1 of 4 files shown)


src/Cache/Internal/ScsDataClient.php
```diff
@@ -186,13 +186,16 @@ public function setLogger(LoggerInterface $logger): void
         $this->logger = $logger;
     }
 
+    /**
+     * @param int|float|null $ttl
+     */
     private function ttlToMillis(?int $ttl = null): int
```
cprice404, line 192: does the argument type need to be updated here? where it says `?int`?

cprice404 (commented):
Thanks for doing this. In general I'm a big +1.

Two minor things:
1. In places where you are removing the `int` type decorator from arg lists etc., would it be possible / desirable to annotate it as `int|float` rather than removing it? I don't have much experience with type annotations in PHP so I will defer to your judgment but wanted to ask.
2. Would like for @pgautier404 to approve too before we merge/release.

GrahamCampbell:
Moved to #187.

PR #187, opened 2024-05-17
feat: Support fractional TTLs via floats/doubles

cprice404:


cprice404:
This is just a dupe of #183 targeted against a branch in the momento org so that the CI can access secrets

PR #191, opened 2024-06-06
feat: adding storage client

pgautier404:


src/Auth/AuthUtils.php
```diff
@@ -47,6 +48,7 @@ public static function parseJwtToken(string $authToken): object
         try {
             $payload = $exploded[1];
             $payload = JWT::jsonDecode(JWT::urlsafeB64Decode($payload));
+            $payload->storage = null;
```
pgautier404, line 51: Explicitly not supporting storage client for legacy tokens unless I hear otherwise. I could regex one of the other endpoints to calculate this, but I'm not sure we want to keep supporting the old tokens?

src/Auth/StringMomentoTokenProvider.php
```diff
@@ -14,13 +14,18 @@ class StringMomentoTokenProvider extends CredentialProvider
...
     protected ?string $trustedCacheEndpointCertificateName = null;
 
     public function __construct(
         string  $authToken,
         ?string $controlEndpoint = null,
         ?string $cacheEndpoint = null,
+        // TODO: adding this arg would be a breaking change for anyone currently passing in
+        //  endpointCertificateName arguments. I am pretty sure that is 0 people, but wanted
+        //  to call it out.
+        ?string $storageEndpoint = null,
```
pgautier404, line 28: Raising a flag here, though as far as I know this wouldn't have any customer impact.

cprice404 (commented):
just clearing this off of my review list for now, please add me again when it's ready for 👀 !

src/Storage/Internal/StorageDataClient.php
```diff
@@ -0,0 +1,263 @@
...
+        } catch (Exception $e) {
+            return ResponseFuture::createResolved(new StorageGetError(new UnknownError($e->getMessage(), 0, $e)));
+        }
+
+        return ResponseFuture::createPending(
+            function () use ($call): StorageGetResponse {
+                try {
+                    $response = $this->processCall($call);
+                } catch (SdkError $e) {
+                    return new StorageGetError($e);
```
pgautier404, line 213: I'll be adding logic here to differentiate between `Element not found` and the rest of the errors. I'll also modify the `StorageGetResponse` and `StorageGetSuccess` objects to add a way to signify that the element was not found.

src/Auth/StringMomentoTokenProvider.php
malandis: I doubt anyone is building this directly as you said.

src/Cache/Errors/Errors.php
```diff
@@ -19,8 +19,21 @@ abstract class MomentoErrorCode
...
+     */
+    public const CACHE_NOT_FOUND_ERROR = "NOT_FOUND_ERROR";
+    /**
+     * Store with specified name doesn't exist
+     */
+    public const STORE_NOT_FOUND_ERROR = "STORE_NOT_FOUND_ERROR";
+    /**
+     * Item with specified name doesn't exist
+     */
+    public const ITEM_NOT_FOUND_ERROR = "ITEM_NOT_FOUND_ERROR";
```
malandis, line 36: We should call this `STORE_ITEM...` to disambiguate vs cache items.

src/Cache/Errors/Errors.php
```diff
@@ -216,13 +229,41 @@ class LimitExceededError extends SdkError
...
+
+/**
+ * Item with specified name doesn't exist
+ */
+class ItemNotFoundError extends SdkError
+{
+    public string $errorCode = MomentoErrorCode::ITEM_NOT_FOUND_ERROR;
+    public string $messageWrapper = 'An item with the specified name does not exist.  To resolve this error, make sure you have created the item before attempting to use it';
+}
+
```
malandis, line 266: Rename should extend here too.

src/Cache/Internal/ScsControlClient.php
```diff
@@ -132,4 +146,56 @@ public function flushCache(string $cacheName): FlushCacheResponse
...
+            $call = $this->grpcManager->client->CreateStore($request);
+            $this->processCall($call);
+        } catch (AlreadyExistsError $e) {
+            return new CreateStoreAlreadyExists();
+        } catch (SdkError $e) {
+            $this->logger->debug("Failed to create store $storeName: {$e->getMessage()}");
+            return new CreateStoreError($e);
+        } catch (\Exception $e) {
+            $this->logger->debug("Failed to create store $storeName: {$e->getMessage()}");
+            return new CreateStoreError(new UnknownError($e->getMessage()));
```
malandis, line 164: Do we not map the error to something more specific than `UnknownError`?

src/Config/Configurations/StorageInRegion.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param ILoggerFactory|null $loggerFactory
+     * @return StorageInRegion
+     */
+    private static function v1(?ILoggerFactory $loggerFactory = null): StorageInRegion
+    {
+        $loggerFactory = $loggerFactory ?? new NullLoggerFactory();
+        $grpcConfig = new StaticGrpcConfiguration(1100);
+        $transportStrategy = new StaticTransportStrategy($grpcConfig, $loggerFactory, self::$maxIdleMillis);
+        return new self($loggerFactory, $transportStrategy);
+    }
```
malandis, line 47: Current guidelines are to only expose `latest` and not `v[N]`

src/Config/Configurations/StorageLaptop.php
```diff
@@ -0,0 +1,45 @@
...
+     * @param ILoggerFactory|null $loggerFactory
+     * @return StorageLaptop
+     */
+    private static function v1(?ILoggerFactory $loggerFactory = null): StorageLaptop
+    {
+        $loggerFactory = $loggerFactory ?? new NullLoggerFactory();
+        $grpcConfig = new StaticGrpcConfiguration(5000);
+        $transportStrategy = new StaticTransportStrategy($grpcConfig, $loggerFactory, self::$maxIdleMillis);
+        return new StorageLaptop($loggerFactory, $transportStrategy);
+    }
```
malandis, line 44: Same here re: `v1`

malandis: Also let's ensure our timeouts are consistent across SDKs

src/Storage/Internal/StorageDataClient.php
```diff
@@ -0,0 +1,248 @@
...
+use Store\_StoreGetRequest;
+use Store\_StorePutRequest;
+use Store\_StoreValue;
+use function Momento\Utilities\validateNullOrEmpty;
+use function Momento\Utilities\validateOperationTimeout;
+use function Momento\Utilities\validateStoreName;
+
+class StorageDataClient implements LoggerAwareInterface
+{
+    private static int $DEFAULT_DEADLINE_MILLISECONDS = 5000;
```
malandis, line 38: TODO: confirm control and data timeouts in configs are consistent across SDKs

src/Storage/Internal/StorageDataClient.php
```diff
@@ -0,0 +1,248 @@
...
+        } catch (Exception $e) {
+            return ResponseFuture::createResolved(new StorageGetError(new UnknownError($e->getMessage(), 0, $e)));
+        }
+
+        return ResponseFuture::createPending(
+            function () use ($call): StorageGetResponse {
+                try {
+                    $response = $this->processCall($call);
+                } catch (ItemNotFoundError $e) {
+                    return new StorageGetSuccess();
```
malandis, line 196: 👍

src/Storage/PreviewStorageClient.php
```diff
@@ -0,0 +1,474 @@
...
+use Momento\Storage\StorageOperationTypes\ListStoresResponse;
+use Momento\Storage\StorageOperationTypes\StorageDeleteResponse;
+use Momento\Storage\StorageOperationTypes\StorageGetResponse;
+use Momento\Storage\StorageOperationTypes\StoragePutResponse;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Preview client to perform operations against the Momento storage service.
+ * WARNING: the API for this client is not yet stable and may change without notice.
```
malandis, line 25: 👍

src/Storage/PreviewStorageClient.php
```diff
@@ -0,0 +1,474 @@
...
+        $this->logger = $logger;
+    }
+
+    /**
+     * Creates a store if it doesn't exist.
+     * @param string $storeName Name of the store to create
+     * @return CreateStoreResponse Represents the result of the create store operation. This result is
+     *  resolved to a type-safe object of one of the following types:<br>
+     *  * CreateStoreSuccess<br>
+     *  * CreateStoreError<br>
```
malandis, line 104: What about store already exists?

src/Storage/PreviewStorageClient.php
```diff
@@ -0,0 +1,474 @@
...
+     * }
+     * </code>
+     */
+    public function putInteger(string $storeName, string $key, int $value): StoragePutResponse
+    {
+        return $this->putIntegerAsync($storeName, $key, $value)->wait();
+    }
+
+    /**
+     * Put the bytes value in the store.
```
malandis, line 309:
```suggestion
     * Put the double value in the store.
```

src/Storage/PreviewStorageClient.php
```diff
@@ -0,0 +1,474 @@
...
+     * we implicitly wait for completion of the request on destruction of the
+     * response future.
+     */
+    public function putDoubleAsync(string $storeName, string $key, float $value): ResponseFuture
+    {
+        return $this->getNextDataClient()->putDouble($storeName, $key, $value);
+    }
+
+    /**
+     * Put the bytes value in the store.
```
malandis, line 337:
```suggestion
     * Put the double value in the store.
```

src/Storage/StorageOperationTypes/StorageOperationTypes.php
```diff
@@ -0,0 +1,581 @@
...
+ */
+class StorageGetError extends StorageGetResponse
+{
+    use ErrorBody;
+
+    public function type(): string
+    {
+        throw $this->innerException();
+    }
+
```
malandis, line 493: Are we throwing exceptions here vs returning `null`?

src/Utilities/_ErrorConverter.php
```diff
@@ -42,11 +45,29 @@ class _ErrorConverter
...
+                    $class = ItemNotFoundError::class;
+                } else {
+                    $class = CacheNotFoundError::class;
+                }
+            } else {
+                $class = self::$rpcToError[$status];
+            }
             return new $class($details, $status, null, $metadata);
+        } else {
+            print("nope");
```
malandis, line 70: debug code? :)

malandis (commented):
No API concerns just some questions

src/Config/Configurations/StorageInRegion.php
pgautier404: This is private so is not exposed.

src/Config/Configurations/StorageLaptop.php
pgautier404: Same here regarding this being private. 

Regarding inconsistency, can you be specific about what the inconsistency is here? Is there a recommended value for this?

src/Storage/PreviewStorageClient.php
pgautier404: Good catch, thanks!

src/Utilities/_ErrorConverter.php
pgautier404: Ha, how did you guess!? Good eye, thanks.

src/Config/Configurations/StorageLaptop.php
malandis: There's not guidance yet. Not a blocker.

src/Storage/StorageOperationTypes/StorageOperationTypes.php
pgautier404: Good question, I wasn't sure how far to take this. Same question applies to `found()` as well, which makes sense as both `null` and `false` in the error class.

src/Cache/Errors/Errors.php
pgautier404: There is no not found error for cache items-- those are represented as misses. This is also how this is defined in the JS SDK. I'm not necessarily opposed to changing this error code to be specific to storage for future-proofing, but we should be consistent across SDKs. I'm going to leave this for now to maintain consistency with other SDKs but an happy to discuss further.

src/Storage/StorageOperationTypes/StorageOperationTypes.php
pgautier404: I switched it so both `type()` and `found()` return `null` from the error class.

malandis (approved):

PR #192, opened 2024-06-21
feat: one time headers

rishtigupta:
## Issue:
https://github.com/momentohq/dev-eco-issue-tracker/issues/894

### DIFF
--- a/src/Cache/Interceptors/AgentInterceptor.php
+++ b/src/Cache/Interceptors/AgentInterceptor.php
@@ -7,17 +7,23 @@ use Grpc\Interceptor;
 
 class AgentInterceptor extends Interceptor
 {
-
-    private const AGENT = "php:0.5.0";
     private bool $isFirstRequest = true;
+    private string $agent;
+    private string $runtimeVersion;
+
+    public function __construct(string $clientType)
+    {
+        $this->agent = sprintf("php:%s:0.5.0", $clientType);
+        $this->runtimeVersion = PHP_VERSION;
+    }
 
     public function interceptUnaryUnary($method, $argument, $deserialize, $continuation, array $metadata = [], array $options = [])
     {
         if ($this->isFirstRequest) {
-            $metadata["agent"] = [self::AGENT];
+            $metadata["agent"] = [$this->agent];
+            $metadata["runtime_version"] = [$this->runtimeVersion];
             $this->isFirstRequest = false;
         }
         return parent::interceptUnaryUnary($method, $argument, $deserialize, $continuation, $metadata, $options);
     }
-
 }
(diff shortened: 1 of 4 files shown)


src/Cache/Interceptors/AgentInterceptor.php
```diff
@@ -7,17 +7,23 @@
...
+    {
+        $this->agent = sprintf("php:%s:0.5.0", $clientType);
+        $this->runtimeVersion = PHP_VERSION;
+    }
 
     public function interceptUnaryUnary($method, $argument, $deserialize, $continuation, array $metadata = [], array $options = [])
     {
         if ($this->isFirstRequest) {
-            $metadata["agent"] = [self::AGENT];
+            $metadata["agent"] = [$this->agent];
```
bruuuuuuuce, line 23: going off what I see here in nodejs

https://github.com/momentohq/client-sdk-javascript/blob/f3fa909eb2da64fe3ba17dc3fd3e5981bd7ce8cf/packages/client-sdk-nodejs/src/internal/pubsub-client.ts#L79-L81

it looks like these should be `Agent` and `Runtime-Version`

rishtigupta: Chris had this PR in Ruby where he mentioned headers must be lower case: https://github.com/momentohq/client-sdk-ruby/commit/606700265113c06c36ea6c603418c606101ddbb6. 
So, I left the headers the way there previously?

bruuuuuuuce: hmmm interesting, either way we should change `runtime_version` -> `runtime-version`. Maybe we need to change the javascript sdk to also be lowercase?

rishtigupta: I can make the runtime_version change. 
@cprice404 Lowercase/uppercase headers?

src/Cache/Interceptors/AgentInterceptor.php
```diff
@@ -7,17 +7,23 @@
...
 {
-
-    private const AGENT = "php:0.5.0";
     private bool $isFirstRequest = true;
+    private string $agent;
+    private string $runtimeVersion;
+
+    public function __construct(string $clientType)
+    {
+        $this->agent = sprintf("php:%s:0.5.0", $clientType);
```
pgautier404, line 16: Can you please change this to 1.10.0 for now so we're closer to the truth? I'm going to set up release please for this repo soon and let it stamp the version into the code, but I'd like to go ahead and update this while we're in here.

src/Cache/Interceptors/AgentInterceptor.php
pgautier404: Looks like lowercase is actually required, so what you have is perfect.

bruuuuuuuce (approved):
:shipit:

PR #198, opened 2024-06-29
chore: update get responses

pgautier404:


cprice404 (approved):
if you have the versioning issues sorted out let's go ahead and merge and release this, and then we can maybe ask Rishti for a few more minutes of review

PR #199, opened 2024-07-03
chore: add release please workflow

pgautier404:


src/Cache/Interceptors/AgentInterceptor.php
```diff
@@ -10,10 +10,12 @@ class AgentInterceptor extends Interceptor
     private bool $isFirstRequest = true;
     private string $agent;
     private string $runtimeVersion;
+    private string $sdkVersion = "1.10.0"; // x-release-please-version
```
rishtigupta, line 13: Do we would need to manually update the sdkVersion whenever we bump the release-please version?

nand4011: I believe release-please knows to update that line based on the `// x-release-please-version`

rishtigupta: Interesting, thanks for letting me know!

rishtigupta (approved):

.github/workflows/on-push-to-main.yaml
```diff
@@ -49,3 +49,19 @@ jobs:
...
+          token: ${{ secrets.MOMENTO_MACHINE_USER_GITHUB_TOKEN }}
+          release-type: php
+          package-name: momentohq/client-sdk-php
+          extra-files: |
+            src/Cache/Interceptors/AgentInterceptor.php
+          changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"chore","section":"Miscellaneous","hidden":false}]'
+
+      - name: Publish
+        run: |
+          echo "Published version: ${{ steps.release.outputs.tag_name }}"
```
nand4011, line 66: Is this to prove it works before adding the actual publishing?

nand4011 (approved):

src/Cache/Interceptors/AgentInterceptor.php
pgautier404: Yep, the `extra-files` option in the gh action tells rp to find and update a version string in this file.

.github/workflows/on-push-to-main.yaml
pgautier404: Publishing a PHP package that's already indexed in packagist is handled by a github webhook that runs on push. When a new version tag is added, it's automatically added to the index.

PR #210, opened 2024-09-24
chore: Run CI on PHP 8.3 and 8.4, fix PHP 8.4 support, pin actions versions

GrahamCampbell:
It looks like the generated code has some compatibility issues with PHP 8.4. For example `_DictionaryGetResponse $response = null` is not allowed without a `?` in front, to explicitly indicate the type is nullable. Would it be permissible to run the codebase through php-cs-fixer to fix compatibility issues post-code-gen?

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -12,9 +12,9 @@ permissions:
 
 jobs:
   dependency-review:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: 'Checkout Repository'
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: 'Dependency Review'
         uses: actions/dependency-review-action@v2
--- a/.github/workflows/on-pull-request.yaml
+++ b/.github/workflows/on-pull-request.yaml
@@ -12,14 +12,14 @@ jobs:
       max-parallel: 1
       fail-fast: true
       matrix:
-        php-version: ["7.4", "8.0", "8.1", "8.2"]
-    runs-on: ubuntu-latest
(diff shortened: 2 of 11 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -12,9 +12,9 @@ permissions:
 
 jobs:
   dependency-review:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
cprice404, line 15: we've had problems in the past with ubuntu 22.04 and some networking issues between github/azure and momento. We can pin these to 24.04 if you like.

.github/workflows/on-pull-request.yaml
```diff
@@ -12,14 +12,14 @@ jobs:
       max-parallel: 1
       fail-fast: true
       matrix:
-        php-version: ["7.4", "8.0", "8.1", "8.2"]
-    runs-on: ubuntu-latest
+        php-version: ['7.4', '8.0', '8.1', '8.2', '8.3', '8.4']
```
cprice404, line 15: definitely in favor of this, thank you.

cprice404 (commented):
thanks! and yes we are happy to run the linter/fixer if it won't break compat with older versions! i'd be happy to add them to our CI jobs and build scripts too to enforce it.

GrahamCampbell:
I've added in the linting to the CI. It can be run locally without `--dry-run` to actually apply the fixes.

.github/workflows/dependency-review.yml
GrahamCampbell: Updated

GrahamCampbell:
Replaced by #211.

PR #211, opened 2024-09-24
chore: Run CI on PHP 8.3 and 8.4, fix PHP 8.4 support, pin actions versions

cprice404:
copy of #210 , rebased onto main


GrahamCampbell (approved):

GrahamCampbell:
I've pushed a commit on the other PR that should fix (at least the first broken part of) the build, by moving the linting phases out to run just once in their own job.

rishtigupta (approved):
:shipit:

PR #214, opened 2024-10-11
feat: add getItemTtl api

rishtigupta:
## PR Description:
- Add `itemGetTtl` API
- Add integration test

### DIFF
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -8,6 +8,7 @@ use Cache_client\_DictionaryGetResponse;
 use Cache_client\_DictionaryIncrementResponse;
 use Cache_client\_GetResponse;
 use Cache_client\_IncrementResponse;
+use Cache_client\_ItemGetTtlResponse;
 use Cache_client\_KeysExistResponse;
 use Cache_client\_ListFetchResponse;
 use Cache_client\_ListLengthResponse;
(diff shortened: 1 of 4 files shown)


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -3328,3 +3329,110 @@ class SetBatchError extends SetBatchResponse
...
+        }
+    }
+
+    /**
+     * @return int|null Remaining time-to-live in milliseconds.
+     */
+    public function remainingTtlMillis(): ?int
+    {
+        return $this->remainingTtlMillis;
+    }
```
malandis, line 3412: Why is the return value nullable?

rishtigupta: oops, that was a mistake from my end

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -3404,7 +3404,7 @@ public function __construct(_ItemGetTtlResponse $response)
     }
 
     /**
-     * @return int|null Remaining time-to-live in milliseconds.
+     * @return int Remaining time-to-live in milliseconds.
      */
     public function remainingTtlMillis(): ?int
```
malandis, line 3409: Also need to change the return type

malandis (approved):

PR #218, opened 2024-10-15
feat: add sorted set put element and fetch by rank

nand4011:
Add sorted set put element.

Add sorted set fetch by rank.

### DIFF
--- a/src/Cache/CacheClient.php
+++ b/src/Cache/CacheClient.php
@@ -45,6 +45,8 @@ use Momento\Cache\CacheOperationTypes\SetIfPresentResponse;
 use Momento\Cache\CacheOperationTypes\SetLengthResponse;
 use Momento\Cache\CacheOperationTypes\SetRemoveElementResponse;
 use Momento\Cache\CacheOperationTypes\SetResponse;
+use Momento\Cache\CacheOperationTypes\SortedSetFetchResponse;
+use Momento\Cache\CacheOperationTypes\SortedSetPutElementResponse;
 use Momento\Cache\CacheOperationTypes\CreateCacheResponse;
 use Momento\Cache\CacheOperationTypes\DeleteCacheResponse;
 use Momento\Cache\CacheOperationTypes\ListCachesResponse;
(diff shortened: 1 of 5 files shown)


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -1,4 +1,6 @@
 <?php
+/** @noinspection PhpIncompatibleReturnTypeInspection */
```
nand4011, line 2: I added these warnings to cut down on the clutter in IntelliJ. I don't think we are likely to miss any issues because of them, but I can remove them if preferred.

src/Cache/CacheClient.php
```diff
@@ -1718,6 +1720,127 @@ public function setRemoveElement(string $cacheName, string $setName, string $ele
...
+     *   $theSet = $response->valueArray();
+     * } elseif ($error = $response->asError()) {
+     *   // handle error condition
+     * }
+     * </code>
+     * If inspection of the response is not required, one need not call wait as
+     * we implicitly wait for completion of the request on destruction of the
+     * response future.
+     */
+    public function sortedSetFetchByRankAsync(string $cacheName, string $sortedSetName, ?int $startRank = 0, ?int $endRank = null, ?int $order = SORT_ASC): ResponseFuture
```
nand4011, line 1808: I'm using the built-in SORT_ASC/SORT_DESC constants for sort order, since our minimum supported php version, 8.0, doesn't support enums. They map to ints. If they supply SORT_ASC or higher, than we choose ascending, otherwise we choose descending. Ascending if they supply null.

src/Cache/Internal/ScsDataClient.php
```diff
@@ -1513,6 +1526,115 @@ function () use ($call): SetRemoveElementResponse {
...
+                $byIndex->setUnboundedStart(new _Unbounded());
+            }
+            if ($endRank) {
+                $byIndex->setExclusiveEndIndex($endRank);
+            } else {
+                $byIndex->setUnboundedEnd(new _Unbounded());
+            }
+            $sortedSetFetchRequest->setByIndex($byIndex);
+
+            if (!$order || $order >= SORT_ASC) {
```
rishtigupta, line 1603: What does >= mean here?

nand4011: order is just an integer. SORT_DESC is 3 and SORT_ASC is 4. So, if the user supplies 3 or lower, we choose descending, or if the supply 4 or higher we choose ascending. I wanted to cover the cases where they gave integers other than SORT_ASC or SORT_DESC.

nand4011: I should change that to `is_null($order)`, since I think 0 will evaluate to ascending with the current logic.

rishtigupta (dismissed):
:shipit:

src/Cache/Internal/ScsDataClient.php
rishtigupta: Maybe in a separate PR (due to CI jobs 😢 )

rishtigupta (approved):
lgtm! :shipit:

Issue #213, opened 2024-10-10
Implement missing SortedSet, TTL, and Scalar APIs in client-sdk-php


rishtigupta:
https://www.notion.so/momentohq/Momento-Replacement-Drop-in-Client-for-phpredis-11bfeb15c97d8078b461fc3ee09260f6

## Sorted set
- [x] sortedSetPutElement
- [x] sortedSetPutElements - Nate
- [x] sortedSetIncrementScore - Michael
- [x] sortedSetRemoveElement - rishti
- [x] sortedSetRemoveElements - Michael
- [x] sortedSetLengthByScore - Nate
- [x] sortedSetGetScore - Michael
- [x] sortedSetFetchByRank
- [x] sortedSetFetchByScore - Nate

## Ttl
- [x] updateTtl
- [x] increaseTtl
- [x] decreaseTtl
- [x] itemGetTtl

## Scalar
- [x] setIfAbsent
- [x] setIfPresent

## Dev docs
- [x] Update doc example snippets (can be sequenced after plugin development)

danielamiao:
@malandis to help complete the doc above for customer review

malandis:
I updated the Notion document with detail not just on the various commands, but specific options and return values as well. See [this section](https://www.notion.so/momentohq/Momento-Replacement-Drop-in-Client-for-phpredis-11bfeb15c97d8078b461fc3ee09260f6?pvs=4#11ffeb15c97d8074b823cf66d8e6b5bf).

malandis:
Nate:
- sortedSetPutElements
- sortedSetFetchByRank

Michael:
- sortedSetGetScore
- sortedSetIncrementScore

malandis:
get score: https://github.com/momentohq/client-sdk-php/pull/221
remove elements: https://github.com/momentohq/client-sdk-php/pull/227
increment score: https://github.com/momentohq/client-sdk-php/pull/229

PR #222, opened 2024-10-16
feat: add sortedSetRemoveElement api

rishtigupta:
## PR Description:
- add `sortedSetRemoveElement` api
- add integ tests

## Issue
https://github.com/momentohq/client-sdk-php/issues/213

### DIFF
--- a/src/Cache/CacheClient.php
+++ b/src/Cache/CacheClient.php
@@ -51,6 +51,7 @@ use Momento\Cache\CacheOperationTypes\SortedSetGetScoreResponse;
 use Momento\Cache\CacheOperationTypes\CreateCacheResponse;
 use Momento\Cache\CacheOperationTypes\DeleteCacheResponse;
 use Momento\Cache\CacheOperationTypes\ListCachesResponse;
+use Momento\Cache\CacheOperationTypes\SortedSetRemoveElementResponse;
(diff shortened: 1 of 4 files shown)


src/Cache/CacheClient.php
```diff
@@ -1904,6 +1905,54 @@ private function getNextDataClient(): ScsDataClient
         return $client;
     }
 
+    /**
+     * Remove an element to a sorted set.
```
malandis, line 1909: TODO: from a sorted set

tests/Cache/CacheClientTest.php
```diff
@@ -3297,6 +3297,59 @@ public function testSortedSetFetchByRankWithNegativeStartRankLessThanEndRank_Thr
         $this->assertEquals(MomentoErrorCode::INVALID_ARGUMENT_ERROR, $response->asError()->errorCode());
     }
 
+    public function testSortedSetRemoveElement_HappyPath()
```
malandis, line 3300: TODO: also test a sorted set with multiple elements:
- remove an element in the sorted set; the sorted set is still there but lacking that element
- remove an element not in the sorted set; the sorted set is the same as before

malandis (commented):
I left some comments for later; let's not blocking merging this. We can address the other issues in a followup PR since CI takes a while.

malandis (approved):

PR #223, opened 2024-10-16
ci: cache extensions

malandis:
Try caching the PHP extensions since they take a long time to
build. We try this both in the lint job and test job.

We also temporarily reduce the number of versions under test in
 order to increase velocity.


nand4011 (approved):
LGTM. Might be worth rerunning the build to see if it's any faster with the cache.

PR #230, opened 2024-10-18
feat: add sortedSetFetchByScore

nand4011:
Add the sorted set fetch by score API, response types, integration tests, and documentation.

### DIFF
--- a/src/Cache/CacheClient.php
+++ b/src/Cache/CacheClient.php
@@ -1899,7 +1899,84 @@ class CacheClient implements LoggerAwareInterface
         return $this->sortedSetFetchByRankAsync($cacheName, $sortedSetName, $startRank, $endRank, $order)->wait();
     }
 
-    // placeholder: sortedSetFetchByScore
+    /**
+     * Fetch the elements in a sorted set by score.
+     *
+     * @param string $cacheName Name of the cache that contains the set.
+     * @param string $sortedSetName The sorted set to fetch.
+     * @param ?float $minScore The minimum score (inclusive) of the
+     * elements to fetch. Defaults to negative infinity.
+     * @param ?float $maxScore The maximum score (inclusive) of the
+     * elements to fetch. Defaults to positive infinity.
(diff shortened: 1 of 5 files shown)


src/Cache/CacheClient.php
```diff
@@ -1899,7 +1899,84 @@ public function sortedSetFetchByRank(string $cacheName, string $sortedSetName, ?
...
+     * Fetch the elements in a sorted set by score.
+     *
+     * @param string $cacheName Name of the cache that contains the set.
+     * @param string $sortedSetName The sorted set to fetch.
+     * @param ?float $minScore The minimum score (inclusive) of the
+     * elements to fetch. Defaults to negative infinity.
+     * @param ?float $maxScore The maximum score (inclusive) of the
+     * elements to fetch. Defaults to positive infinity.
+     * @param ?int $order The order to fetch the elements in. Defaults to Ascending.
+     *  Will be treated as ascending if SORT_ASC or higher is supplied, and descending if SORT_DESC or lower is supplied.
```
malandis, line 1912: What does "or higher" and "or lower" mean here?

src/Cache/Internal/ScsDataClient.php
```diff
@@ -1701,6 +1702,89 @@ function () use ($call): SortedSetFetchResponse {
...
+
+            if (is_null($offset)) {
+                $byScore->setOffset(0);
+            } else {
+                $byScore->setOffset($offset);
+            }
+
+            if (is_null($count)) {
+                // a negative count returns all elements.
+                $byScore->setCount(-1);
```
malandis, line 1746: What about count 0?

nand4011: Good catch.

malandis (approved):

src/Cache/CacheClient.php
nand4011: order is an integer. SORT_ASC is 4 and SORC_DESC is 3. you can supply any integer value. We're saying that if you supply 3 or less it is descending and if you supply 4 or more it is ascending.

PR #233, opened 2024-10-18
fix: stop using is_string() for validation

nand4011:
Stop using is_string() for validation as it fails for strings that have automatically been converted into integers when used as keys in associative arrays.

Update isNullOrEmpty to check if the given value is a scalar (a string, int, float, or bool) and then check if its string value is empty instead of explicitly taking a string.

Remove a dictionary set fields test case that checks for an error when given this array: [""]. PHP will implicitly treat it as this associative array: [0 => ""], which is valid with our new validation. We need to choose whether to support integer string keys or whether to be able to reject this specific type of invalid input.

### DIFF
--- a/tests/Cache/CacheClientTest.php
+++ b/tests/Cache/CacheClientTest.php
@@ -2119,10 +2119,6 @@ class CacheClientTest extends TestCase
     public function testDictionarySetFieldsWithEmptyItems_IsError()
     {
         $dictionaryName = uniqid();
-        $items = [""];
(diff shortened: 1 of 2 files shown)


tests/Cache/CacheClientTest.php
```diff
@@ -2119,10 +2119,6 @@ public function testDictionarySetFieldsWithNullItems_ThrowsException()
     public function testDictionarySetFieldsWithEmptyItems_IsError()
     {
         $dictionaryName = uniqid();
-        $items = [""];
```
nand4011, line 2122: Unfortunately I don't know how to make this an invalid input while also allowing integer string keys.

malandis (approved):

rishtigupta (approved):
:shipit:

PR #234, opened 2024-10-18
feat: add sortedSetLengthByScore

nand4011:
Add the sorted set length by score API, integration tests, and documentation.

### DIFF
--- a/src/Cache/CacheOperationTypes/CacheOperationTypes.php
+++ b/src/Cache/CacheOperationTypes/CacheOperationTypes.php
@@ -22,10 +22,10 @@ use Cache_client\_SetContainsResponse;
 use Cache_client\_SetFetchResponse;
 use Cache_client\_SetLengthResponse;
 use Cache_client\_SortedSetFetchResponse;
+use Cache_client\_SortedSetLengthByScoreResponse;
 use Cache_client\ECacheResult;
 use Closure;
 use Control_client\_ListCachesResponse;
-use Generator;
 use Momento\Cache\Errors\SdkError;
 use Momento\Cache\Errors\UnknownError;
 use Throwable;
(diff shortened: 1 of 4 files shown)


src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -3239,7 +3239,83 @@ class SetRemoveElementError extends SetRemoveElementResponse
...
+ */
+class SortedSetLengthByScoreSuccess extends SortedSetLengthByScoreResponse
+{
+    private int $length;
+
+    public function __construct(_SortedSetLengthByScoreResponse $response)
+    {
+        parent::__construct();
+        $this->length = $response->getFound() ? $response->getFound()->getLength() : 0;
+    }
```
malandis, line 3296: Just leaving a note: the "or 0" case should never happen. I opted to keep the protobuf responses out of here and pass concrete values to here from the data client.

src/Cache/CacheOperationTypes/CacheOperationTypes.php
```diff
@@ -3239,7 +3239,83 @@ class SetRemoveElementError extends SetRemoveElementResponse
...
+     * @return SortedSetLengthByScoreError|null Returns the error subtype if the request returned an error and null otherwise.
+     */
+    public function asError(): ?SortedSetLengthByScoreError
+    {
+        if ($this->isError()) {
+            return $this;
+        }
+        return null;
+    }
+}
```
malandis, line 3283: Length is the hit/miss/error pattern

malandis (commented):
In the JS SDK for the collections, length is hit/miss/error

src/Cache/CacheOperationTypes/CacheOperationTypes.php
nand4011: Updated.

malandis (approved):

PR #235, opened 2024-10-18
chore: make order in the sorted set functions more explicit

nand4011:
change the order argument in the sorted set functions to explicitly take SORT_ASC and SORT_DESC, and treat other values as ascending. This replaces the previous logic where SORT_DESC or lower was descending, and SORT_ASC or higher was ascending. Using built-in constants in a comparison was confusing.


malandis (approved):
I'd make “any other values “ an error. We can change that later.

PR #238, opened 2024-10-21
feat: add flags for toggling inclusivity on fetchByScore scores

nand4011:
Add $inclusiveMin and $inclusiveMax to sortedSetFetchByScore to allow callers to toggle whether the min and max scores should be inclusive or exclusive. They default to inclusive.

Add tests for the flags.

Change the behavior for an unrecognized sort order from defaulting to ascending to returning an error.

Make $order not nullable. There isn't a reason for it to be null and it has a default argument.

### DIFF
--- a/src/Utilities/_DataValidation.php
+++ b/src/Utilities/_DataValidation.php
@@ -202,7 +202,7 @@ if (!function_exists('validateSortedSetElements')) {
                 throw new InvalidArgumentError("Sorted set score must be a float");
             }
 
-            validateValueName($value);
+            validateNullOrEmpty($value, "Sorted set value");
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Utilities/_DataValidation.php
```diff
@@ -233,3 +233,12 @@ function validateSortedSetScores(?float $minScore, ?float $maxScore): void
         }
     }
 }
+
+if (!function_exists('validateSortedSetOrder')) {
+    function validateSortedSetOrder(int $order): void
+    {
+        if ($order != SORT_ASC && $order != SORT_DESC) {
```
rishtigupta, line 240: 💪

rishtigupta (approved):
:shipit:

PR #239, opened 2024-10-21
chore: rename the ThrowsException tests to ReturnsError

nand4011:
Rename the tests that are named ThrowsException to ReturnsError, since none of them test a condition that throws an exception.


rishtigupta (approved):
:shipit:

malandis (approved):

PR #242, opened 2024-10-25
chore: remove the type from sorted set scores so they support ints

nand4011:
Remove the type that the score arguments use in sorted set methods. When they were explicitly float, it was an error to pass in an int score. They are now defined as float|int, but are left untyped in the functions because union types are unsupported in the oldest versions of PHP we support.

Update data validation to handle the new potential score types.

Rearrange the arguments of sorted set functions that take a score range to put the min/max scores and min/max inclusive flags next to each other instead of interleaving them.

### DIFF
--- a/src/Utilities/_DataValidation.php
+++ b/src/Utilities/_DataValidation.php
@@ -198,9 +198,7 @@ if (!function_exists('validateSortedSetElements')) {
     function validateSortedSetElements(array $elements): void
     {
         foreach ($elements as $value => $score) {
-            if (!is_float($score)) {
-                throw new InvalidArgumentError("Sorted set score must be a float");
-            }
+            validateSortedSetScore($score);
 
             validateNullOrEmpty($value, "Sorted set value");
         }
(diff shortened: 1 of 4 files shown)


src/Utilities/_DataValidation.php
```diff
@@ -198,9 +198,7 @@ function validateSortedSetRanks(?int $startRank, ?int $endRank): void
     function validateSortedSetElements(array $elements): void
     {
         foreach ($elements as $value => $score) {
-            if (!is_float($score)) {
-                throw new InvalidArgumentError("Sorted set score must be a float");
-            }
+            validateSortedSetScore($score);
```
malandis, line 201: 👍

malandis (commented):
Also need to validate that the scores passed to fetch by score or length by score are either null, int or float (unlike the ones where the score is not nullable)

src/Utilities/_DataValidation.php
nand4011: fetchByScore and lengthByScore already call validateSortedSetScores, which I modified to validate the types.

rishtigupta (approved):
:shipit:

PR #244, opened 2024-10-25
chore: add named argument sorted-set example

rishtigupta:


examples/sorted-set-example.php
```diff
@@ -94,6 +94,18 @@ function printBanner(string $message, LoggerInterface $logger): void
     exit(1);
 }
 
+// fetch sorted set by score (0 to 2) in descending order
+$logger->info("Fetching sorted set $SET_NAME by score in descending order\n");
+$response = $client->sortedSetFetchByScore($CACHE_NAME, $SET_NAME, $minScore = 0, $maxScore = 2, $inclusiveMin = true, $inclusiveMax = true, $order=SORT_DESC);
```
nand4011, line 99: If you're using named arguments, can't you leave off $inclusiveMin and $inclusiveMax?

rishtigupta: I thought we want to show named arguments for those too?

nand4011: I was thinking that we would use named arguments to show that you could change sort order without having to specify all of the other arguments, assuming you were using the defaults.

nand4011, line 99:
```suggestion
$response = $client->sortedSetFetchByScore($CACHE_NAME, $SET_NAME, minScore: 0, maxScore: 2, order: SORT_DESC);
```

nand4011 (approved):

PR #246, opened 2024-11-04
feat: add ReadConcern config

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1055 

Adds the `read-concern` header with `BALANCED` and `CONSISTENT` options to the cache configs.

Tested manually and added a test case for ensuring the read concern header can be overridden. 
Also switched the cache test client to use the consistent header to match the JS sdk, but can revert that if not needed here.

### DIFF
--- a/src/Config/Configurations/InRegion.php
+++ b/src/Config/Configurations/InRegion.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 namespace Momento\Config\Configurations;
 
 use Momento\Config\Configuration;
+use Momento\Config\ReadConcern;
 use Momento\Config\Transport\StaticGrpcConfiguration;
 use Momento\Config\Transport\StaticStorageTransportStrategy;
 use Momento\Logging\ILoggerFactory;
@@ -40,6 +41,7 @@ class InRegion extends Configuration
         $loggerFactory = $loggerFactory ?? new NullLoggerFactory();
         $grpcConfig = new StaticGrpcConfiguration(1100);
         $transportStrategy = new StaticStorageTransportStrategy($grpcConfig, $loggerFactory, self::$maxIdleMillis);
-        return new self($loggerFactory, $transportStrategy);
+        $readConcern = ReadConcern::BALANCED;
+        return new self($loggerFactory, $transportStrategy, $readConcern);
     }
 }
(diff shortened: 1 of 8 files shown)


src/Config/Configurations/Laptop.php
```diff
@@ -38,6 +39,7 @@ public static function v1(?ILoggerFactory $loggerFactory = null): Laptop
         $loggerFactory = $loggerFactory ?? new NullLoggerFactory();
         $grpcConfig = new StaticGrpcConfiguration(5000);
         $transportStrategy = new StaticStorageTransportStrategy($grpcConfig, $loggerFactory, self::$maxIdleMillis);
-        return new Laptop($loggerFactory, $transportStrategy);
+        $readConcern = ReadConcern::BALANCED;
+        return new Laptop($loggerFactory, $transportStrategy, $readConcern);
```
malandis, line 43: Is this necessary given that the constructor defaults to read concern to null (balanced)?

src/Config/Configurations/InRegion.php
```diff
@@ -40,6 +41,7 @@ public static function v1(?ILoggerFactory $loggerFactory = null): InRegion
         $loggerFactory = $loggerFactory ?? new NullLoggerFactory();
         $grpcConfig = new StaticGrpcConfiguration(1100);
         $transportStrategy = new StaticStorageTransportStrategy($grpcConfig, $loggerFactory, self::$maxIdleMillis);
-        return new self($loggerFactory, $transportStrategy);
+        $readConcern = ReadConcern::BALANCED;
+        return new self($loggerFactory, $transportStrategy, $readConcern);
```
malandis, line 45: Same question here re: defaults

malandis (commented):
Nit: re instantiating the in region and laptop configs

src/Config/Configurations/Laptop.php
anitarua: It's not necessary, but I added it for clarity. Can remove if you think that would be better.

I'd been referencing the [JS sdk](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/config/configurations.ts#L76), which included it in the configs for documentation I think, but not actually necessary when computing whether to send the header (references: [node](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/cache-data-client.ts#L245-L248) and [web](https://github.com/momentohq/client-sdk-javascript/blob/44be3bfbfbe5ff2c562cfcb7a4cb12165de6e676/packages/client-sdk-web/src/internal/client-metadata-provider.ts#L43-L46)).

malandis (approved):

PR #247, opened 2024-11-05
chore: improve resource exhausted error message

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1042

When a RESOURCE_EXHAUSTED error is received, we now parse the err metadata in order to report the cause in the message wrapper. If the metadata is not available, we try to use string matching on the error details.


rishtigupta (approved):
:shipit:

PR #251, opened 2024-11-21
chore: add Lambda configuration with values ported from nodejs

pgautier404:


src/Config/Configurations/Lambda.php
```diff
@@ -0,0 +1,44 @@
...
+     * Provides the version 1 configuration for a Lambda environment. This configuration is guaranteed not to change
+     * in future releases of the SDK.
+     * @param ILoggerFactory|null $loggerFactory
+     * @return Lambda
+     */
+    public static function v1(?ILoggerFactory $loggerFactory = null): Lambda
+    {
+        $loggerFactory = $loggerFactory ?? new NullLoggerFactory();
+        $grpcConfig = new StaticGrpcConfiguration(1100);
+        $transportStrategy = new StaticStorageTransportStrategy($grpcConfig, $loggerFactory, self::$maxIdleMillis);
```
anitarua, line 40: iirc, the main difference in the lambda config was that keepalives were turned off. should we also add the keepalive setting to the grpc config and turn them off here?

nand4011: Yeah, the keepalives were causing errors with infrequently invoked lambdas. If you want to add that in a future pr, we should get rid of the v1 config for now and only use latest.

pgautier404: Good to know, thanks. I'll add that in.

pgautier404: Ok, I added the keepalive params into the gRPC config before realizing that keepalive is opt-in. The nodejs `Lambda` configuration just omits the keepalive parameters to use the gRPC defaults. So the `Lambda` config is good to go as-is and we're free to add keepalives to new configs and new versions of the existing configs in the future as we see fit.

anitarua (approved):
lgtm!

I was going to suggest just updating the keepalive settings in the other configs for now, but I guess that could impact existing customers and maybe would be better to do in a `fix` or `feat` commit.

pgautier404:
> lgtm!
> 
> I was going to suggest just updating the keepalive settings in the other configs for now, but I guess that could impact existing customers and maybe would be better to do in a `fix` or `feat` commit.

Yep, that was exactly what I was thinking.

PR #253, opened 2025-03-01
feat: adding SortedSetUnionStore operation

pgautier404:


rishtigupta:
LGTM!

src/Cache/CacheClient.php
```diff
@@ -2279,6 +2281,89 @@ public function sortedSetLengthByScore(string $cacheName, string $sortedSetName,
         return $this->sortedSetLengthByScoreAsync($cacheName, $sortedSetName, $minScore, $maxScore, $inclusiveMin, $inclusiveMax)->wait();
     }
 
+    /**
+     * Computes the union of all the source sets and stores the result in itself. If the set does not
```
malandis, line 2285:
```suggestion
     * Computes the union of all the source sets and stores the result in the destination key. If the set does not
```

src/Cache/CacheClient.php
```diff
@@ -2279,6 +2281,89 @@ public function sortedSetLengthByScore(string $cacheName, string $sortedSetName,
...
+     * exist, it is created with the given `ttl`. If it exists, it is overwritten with the result and
+     * its ttl is set to the given `ttl`. If the set exists but the result of the union is empty, it is deleted.
+     * The union is computed by applying the corresponding weight multiplier to the score of all elements
```
malandis, line 2288: The `destination` key doesn't actually have to be a sorted set. It could be or not be any other type; it'll be overwritten.

src/Cache/Internal/ScsDataClient.php
```diff
@@ -2031,6 +2038,75 @@ function () use ($call, $value): SortedSetGetScoreResponse {
...
+    ): ResponseFuture
+    {
+        try {
+            $collectionTtl = $this->returnCollectionTtl($ttl);
+            // The number of source sets is currently limited to 2. I'm just adding this validation and
+            // some validation of the contents of $sources here for now, but will move to _DataValidation
+            // eventually.
+            if (count($sources) > 2) {
+                throw new InvalidArgumentError("The number of source sets is currently limited to 2");
+            }
```
malandis, line 2059: Are we leaving this at <= 2 for this PR or did you intend to remove this?

malandis (commented):
Couple small issues otherwise 👍

src/Cache/Internal/ScsDataClient.php
pgautier404: This is an explicit service limit, so I was planning to keep this code and comment around until we can know the best way to validate this more formally.

tests/Cache/CacheClientTest.php
```diff
@@ -3784,6 +3785,46 @@ public function testSortedSetLengthByScoreWithMinScoreLargerThanMaxScore_Returns
...
+            60
+        );
+        $this->assertNull($response->asError());
+        $this->assertNotNull($response->asSuccess(), "Expected a success but got: $response");
+        $this->assertEquals(6, $response->asSuccess()->length());
+
+        $this->client->sortedSetFetchByScore($this->TEST_CACHE_NAME, $sortedSetName3);
+        $this->assertNull($response->asError());
+        $this->assertNotNull($response->asSuccess(), "Expected a hit but got: $response");
+        $this->assertEquals(6, $response->asSuccess()->length());
```
malandis, line 3825: Can you also verify the scores are as expected, add a test for the other two aggregation variants, and one for different weightings as a sanity check?

pgautier404: I am planning on adding a ton of testing as a followup but want to go ahead and get this approved. I still need to cut a release to use in the `momento-php-redis-client`.

malandis (approved):

PR #256, opened 2025-03-04
chore: adding more testing for sorted set union store

pgautier404:
This commit adds a number of integration tests for the sorted set union store API. It also corrects the handling of the aggregate function when users explicitly pass in `null` for that argument. The PHP Redis client library is **not** affected by the existing SDK logic because it validates the correctness of the aggregate function arguments before making any SDK calls.


nand4011 (approved):

rishtigupta (approved):
:shipit:

PR #259, opened 2025-04-22
chore: remove the storage client

nand4011:
Delete the preview storage client and associated configuration and tests.

Update to the newest version of the protos that doesn't have storage support.


pgautier404 (approved):
:shipit:

PR #271, opened 2026-07-06
fix: convert non-OK batch stream statuses into SDK errors

GrahamCampbell:
`getBatch` never checks the call's terminal status, so a stream ending with a non-OK grpc-status is reported as a successful batch containing only the items that arrived before the failure. `setBatch` checks it but wraps every failure in a generic `UnknownError`. Both methods now convert non-OK statuses through `_ErrorConverter` into the typed SDK errors that unary calls produce.


anitarua:
moved this commit to a branch on the repo and merging https://github.com/momentohq/client-sdk-php/pull/272 instead due to the usual issue (env vars not available to CI from forks for security reasons)

